PR #89, opened 2019-03-14
Principal with Key Management

fredfortier:
Notes:
- Deploying the Enigma contract does not work. You'll notice that related sections of the code are commented out. I did not completely remove them because we may want to re-enable them or adapt to your integration test framework. There are two reasons why I chose to disable this: 
  1. Rust Web3 does not support deploying contracts with linked libraries: https://github.com/tomusdrw/rust-web3/issues/186. We could submit a PR to enable it but I determined that it was not a priority.
  2. New business logic like the worker selection algorithm requires extensive setup (registering multiple workers, crossing epoch, etc).

  Here is what I did to compensate:
  1. Added lower level unit tests with mock data that does not interface with the live contract.
  2. Made the CLI commands more granular (which useful beyond testing) and added a mode to `Enigma.specs.js` that invokes the live Principal node. This way, we can substitute the mock for live Principal node and without changing the tests. This was the most convenient way for me to test and also a good bridge to the full integration test script.
- Added a `confirmations` parameter to the `enigma_contract` module in addition to supporting the new methods in the contract.
- I'm using a specific terminology for some of the for the `epoch` and `keys` modules. I'm also using the `_u` and `_t` but there was still overlap: 1) Provider: Provide external data; 2) Keeper: Store and verify data in the enclave. I'm happy to refactor this if you suggest a different convention.
- The `keys_keeper_types` module is copied across: `app/src/esgx/keys_keeper_u.rs` and `enclave/src/keys_keeper_t/keeper_types_t.rs`. I tried to make a reusable module that works both in the enclave and out ran into issues and aborted for now.
- I created `enigma-tools-t/src/document_storage_t.rs` which is based upon `storage_t.rs` but supports variable byte array size. It also has improved error handling. I believe that it can replace `storage_t.rs`.

TODOs:
- Seal the epoch data. I have a plan for this to did not have time to finish. Instead of sealing the entire epoch data, I want to seal the nonce + a hash. The `epoch_provider` will fetch the epoch data and call `set_worker_params` and pass the nonce. Instead of incrementing the nonce, the `epoch_keeper` will verify the data against the hash and store in memory.
- I could not get the Makefile to copy `enigma-types.h` in the enclave folder. I included a temporary workaround on line 111 of `enigma-principal/Makefile`. I would appreciate if you could take a look. This causes the build to fail on the CI server.
- Worker selection is disabled for the time being (it always return the keys). I'm working on testing more thoroughly and will push an update shortly.
- I cherry picked commit 346a8e40827ee4f9af0f56186d72ac916f26cf6f because I needed a working SPID. Now the `attestation_service::service::test` tests are failing. The root cause is not obvious to me. I will wait until the related PR is finalized to troubleshoot.
- Logging is not implemented in the enclave.


eng-wasm/dispatch/src/lib.rs
```diff
@@ -0,0 +1,256 @@
+#![recursion_limit="128"]
```
elichai: What's the deal with the `eng-wasm/dispatch`?  I think this is something from a merge got wrong.

enigma-core/app/enigma-core-app.iml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elichai: Please add the `iml`  files to the `.gitignore`

enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn find_epoch_contract_addresses(reader: PrincipalMessageReader, sig: [u8; 65], epoch_state: EpochState) -> Result<Vec<H256>, Error> {
+        let worker = reader.get_signing_address(sig)?;
+        let addrs = epoch_state.get_contract_addresses(&worker)?;
+        Ok(addrs)
+    }
+
+    #[logfn(DEBUG)]
+    fn get_state_keys_internal(epoch_provider: Arc<EpochProvider>, request: StateKeyRequest, eid: sgx_enclave_id_t) -> Result<Value, Error> {
+        println!("Got get_state_keys request: {:?}", request);
+        let reader = PrincipalMessageReader::new(request.data.clone().into())?;
```
elichai: the `Into` here contains an unwrap. this means that if you receive a non hex string in the JSON-RPC the principal node will crash.
I think the principal node should be a lot more robust, at least in the interface side.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -0,0 +1,267 @@
...
+use enigma_tools_u::web3_utils::enigma_contract::{ContractFuncs, ContractQueries, EnigmaContract};
+use epoch_u::epoch_types::{ConfirmedEpochState, EpochState, WorkersParameterizedEvent};
+use esgx::epoch_keeper_u::set_worker_params;
+use esgx::general::{ENCLAVE_DIR, storage_dir};
+use keys_u::keeper_types_u::InputWorkerParams;
+
+pub struct EpochProvider {
+    pub contract: Arc<EnigmaContract>,
+    pub epoch_state: Arc<Mutex<Option<EpochState>>>,
+    pub eid: Arc<AtomicU64>,
```
elichai: I think the `eid` can be regular `u64` and doesn't need to be `AtomicU64` because there's no need to it to be atomic because it should never change.

enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -0,0 +1,171 @@
...
+            let request = params.parse::<StateKeyRequest>()?;
+            let eid = epoch_provider.eid.load(Ordering::SeqCst);
+            let body = match Self::get_state_keys_internal(epoch_provider, request, eid) {
+                Ok(body) => body,
+                Err(err) => return Err(ServerError { code: ErrorCode::InternalError, message: format!("Unable to get keys: {:?}", err), data: None }),
+            };
+            Ok(body)
+        });
+        let server = ServerBuilder::new(io)
+            .cors(DomainsValidation::AllowOnly(vec![AccessControlAllowOrigin::Null]))
```
elichai: What is the need for CORS headers?

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -63,117 +117,196 @@ impl PrincipalManager {
...
-    fn new_delegated(config: PrincipalConfig, contract: EnigmaContract, eid: sgx_enclave_id_t) -> Self {
-        let as_service = service::AttestationService::new(&config.attestation_service_url);
-        PrincipalManager { eid, config, as_service, contract }
-    }
+    fn get_signing_address(&self) -> Result<String, Error> { Ok(self.report_manager.get_signing_address()?) }
 
     fn get_contract_address(&self) -> Address { self.contract.address() }
 
-    fn get_quote(&self) -> Result<String, Error> { Ok(retry_quote(self.eid, &self.config.spid, 18)?) }
+    fn get_account_address(&self) -> Address { self.contract.account.clone() }
```
elichai: `Address` Implements `Copy` so no need to clone it.

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -63,117 +117,196 @@ impl PrincipalManager {
...
-    }
+    fn get_network_url(&self) -> String { self.config.url.clone() }
 
-    fn get_signing_address(&self) -> Result<String, Error> {
-        let mut signing_address = esgx::equote::get_register_signing_address(self.eid)?;
-        // remove 0x
-        signing_address = signing_address[2..].to_string();
-        Ok(signing_address)
+    fn get_block_number(&self) -> Result<U256, Error> {
+        let block_number = self.get_web3().eth().block_number().wait().unwrap();
```
elichai: Why do you unwrap if you return a result?

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -63,117 +117,196 @@ impl PrincipalManager {
...
+        Ok(block_number)
     }
 
-    fn get_account_address(&self) -> Address { self.contract.account }
+    fn register<G: Into<U256>>(&self, signing_address: String, gas_limit: G) -> Result<H256, Error> {
+        let registration_params = self.report_manager.get_registration_params()?;
+        let report = registration_params.report.clone();
+        let signature = registration_params.signature.clone();
+        let receipt = self.contract.register(signing_address, report, signature, gas_limit, self.config.confirmations as usize)?;
```
elichai:
```suggestion
        let receipt = self.contract.register(signing_address, registration_params.report, registration_params.signature, gas_limit, self.config.confirmations as usize)?;
```
You can pass the report and signature directly without cloning them in memory

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -63,117 +117,196 @@ impl PrincipalManager {
...
         // get enigma contract
         let enigma_contract = &self.contract;
-        let gas_limit: U256 = gas_limit.into();
-        // register worker
-        //0xc44205c3aFf78e99049AfeAE4733a3481575CD26
-        let signer = self.get_signing_address()?;
-        println!("signing address = {}", signer);
-        enigma_contract.register(&signer, &rlp_encoded, gas_limit)?;
+        // Start the WorkerParameterized Web3 log filter
+        let eid: Arc<AtomicU64> = Arc::new(AtomicU64::new(self.eid));
```
elichai: Again, the eid can stay u64 (or more clearly `sgx_enclave_id_t`)

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -183,67 +316,62 @@ mod test {
     /// The testing is looking for atleast 2 emmits of the WorkersParameterized event and compares the event triggerd
     /// If the event name is different or if it takes more than 30 seconds then the test will fail.
     #[test]
-    //#[ignore]
+    #[ignore]
     fn test_full_principal_logic() {
```
elichai, line 335: Why is this test commented out?
I think it's a pretty important test, unless there's a different test that checks if random emission with simulated blocks

enigma-principal/app/src/cli/app.rs
```diff
@@ -1,79 +1,101 @@
...
-
-pub use esgx::general::ocall_get_home;
+use epoch_u::epoch_provider::EpochProvider;
+use esgx::general::{ENCLAVE_DIR, storage_dir};
 
+#[logfn(INFO)]
 pub fn start(eid: sgx_enclave_id_t) -> Result<(), Error> {
     let opt = cli::options::Opt::from_args();
-    let config = deploy_scripts::load_config(opt.deploy_config.as_str())?;
+    let _config = deploy_scripts::load_config(opt.deploy_config.as_str())?;
```
elichai: If you don't use this config why not remove it?

enigma-principal/app/src/cli/app.rs
```diff
@@ -1,79 +1,101 @@
...
+    let report_manager = ReportManager::new(principal_config.clone(), eid)?;
+    let signing_address = report_manager.get_signing_address()?;
 
     if opt.info {
-        cli::options::print_info(&sign_key);
+        cli::options::print_info(&signing_address);
+    } else if opt.sign_address {
+        let mut path = storage_dir();
+        path.join(ENCLAVE_DIR);
+        path.push("principal-sign-addr.txt");
```
elichai, line 25: Why do you save the signing address to a file?

enigma-tools-t/src/document_storage_t.rs
```diff
@@ -0,0 +1,157 @@
+use sgx_tseal::SgxSealedData;
+#[cfg(not(target_env = "sgx"))]
```
elichai: You can remove this.
It's something old that we don't use.

enigma-tools-t/src/document_storage_t.rs
```diff
@@ -0,0 +1,157 @@
...
+        to_sealed_log(&sealed_data, sealed_log, sealed_log_size as u32);
+        Ok(())
+    }
+
+    /// Unseal sealed log
+    /// param: sealed_log_in : the encrypted blob
+    pub fn unseal(sealed_log_in: &mut [u8]) -> Result<Option<Self>, EnclaveError> {
+        let sealed_log_size: usize = SEAL_LOG_SIZE;
+        let sealed_log = sealed_log_in.as_mut_ptr();
+        let sealed_data = match from_sealed_log::<Self>(sealed_log, sealed_log_size as u32) {
```
elichai, line 48: I think this should get a deeper look to see if we can use the "safer" ` unseal_data()`

enigma-tools-t/src/document_storage_t.rs
```diff
@@ -0,0 +1,157 @@
...
+        let sealed_data = SgxSealedData::<Self>::seal_data_ex(
+            0x0001, //key policy
+            attribute_mask,
+            0, //misc mask
+            &additional,
+            &self,
+        )?;
+        let sealed_log = sealed_log_out.as_mut_ptr();
+        let sealed_log_size: usize = SEAL_LOG_SIZE;
+        to_sealed_log(&sealed_data, sealed_log, sealed_log_size as u32);
```
elichai, line 39: Is that really needed? can we just use the Self this returns?
I'll prefer if we didn't use the unsafe functions here.

enigma-tools-t/src/document_storage_t.rs
```diff
@@ -0,0 +1,157 @@
...
+    use super::*;
+
+//use std::untrusted::fs::*;
+
+    /* Test functions */
+    pub fn test_document_sealing_storage() {
+        // generate mock data
+        let mut doc: SealedDocumentStorage<[u8; 32]> = SealedDocumentStorage {
+            version: 0x1234,
+            data: [0; 32],
```
elichai:
```suggestion
            data: [b'i'; 32],
```

elichai: Then you can remove the for loop

enigma-tools-t/src/document_storage_t.rs
```diff
@@ -0,0 +1,157 @@
...
+use sgx_types::marker::ContiguousMemory;
+use std::io::{Read, Write};
+use std::path::PathBuf;
+use std::string::*;
+use std::untrusted::fs;
+use std::untrusted::fs::{File, remove_file};
+
+use common::errors_t::EnclaveError;
+
+pub const SEAL_LOG_SIZE: usize = 2048;
```
elichai, line 12: Why is the whole 2048?
What if you sealed a `[u8; 8096]`?

enigma-tools-t/Cargo.toml
```diff
@@ -1,3 +1,5 @@
+cargo-features = ["rename-dependency"]
```
elichai: Why do you need this? I don't see any dependency that you're renaming

enigma-principal/enclave/src/lib.rs
```diff
@@ -59,50 +74,63 @@ fn get_sealed_keys_wrapper() -> asymmetric::KeyPair {
...
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn ecall_get_enc_state_keys(msg: *const u8, msg_len: usize,
+                                                  addrs: *const u8, addrs_len: usize, sig: &[u8; 65],
+                                                  enc_response_out: *mut u8, enc_response_len: &mut usize,
+                                                  sig_out: &mut [u8; 65]) -> EnclaveReturn {
+    let msg_bytes = slice::from_raw_parts(msg, msg_len).to_vec();
+    let addrs_bytes = slice::from_raw_parts(addrs, addrs_len).to_vec();
+    let response = match ecall_get_enc_state_keys_internal(msg_bytes, addrs_bytes, sig.clone(), sig_out) {
```
elichai: why clone the signature?

enigma-principal/app/src/esgx/keys_keeper_u.rs
```diff
@@ -0,0 +1,118 @@
...
+/// let sig = StringWrapper("43f19586b0a0ae626b9418fe8355888013be1c9b4263a4b3a27953de641991e936ed6c4076a2a383b3b001936bf0eb6e23c78fbec1ee36f19c6a9d24d75e9e081c".to_string());
+/// let request = StateKeyRequest { data: msg, sig: sig };
+/// let response = get_enc_state_keys(enclave.geteid(), request, None).unwrap();
+/// ```
+#[logfn(DEBUG)]
+pub fn get_enc_state_keys(eid: sgx_enclave_id_t, request: StateKeyRequest, epoch_addrs: Option<Vec<H256>>) -> Result<StateKeyResponse, Error> {
+    let mut retval: EnclaveReturn = EnclaveReturn::Success;
+    let mut sig_out: [u8; 65] = [0; 65];
+    let mut enc_response = vec![0u8; MAX_ENC_RESPONSE_LEN];
```
elichai: What if it's higher than 100_000?
you should use: https://github.com/enigmampc/enigma-core-internal/blob/develop/enigma-core/enclave/src/ocalls_t.rs#L26
And then you unbox the memory, see this for reference: 
https://github.com/enigmampc/enigma-core-internal/blob/develop/enigma-core/app/src/km_u.rs#L65

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    for addr in sc_addrs {
+        let mut key: Option<StateKey> = match guard.get(addr) {
+            Some(key) => {
+                let mut buf: StateKey = [0; 32];
+                buf.copy_from_slice(&key[..]);
+                Some(buf)
+            }
+            None => None,
+        };
+        if key.is_none() {
```
elichai: you can put in `Some(key) => results.push(key.clone())` and in None => everything you did in `if key.is_none()`

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        results.push(key);
+    }
+    Ok(results)
+}
+
+/// Creates new state keys both in the cache and as sealed documents
+fn new_state_keys(guard: &mut SgxMutexGuard<HashMap<ContractAddress, StateKey, RandomState>>, sc_addrs: &Vec<ContractAddress>) -> Result<Vec<StateKey>, EnclaveError> {
+    let mut results: Vec<StateKey> = Vec::new();
+    for addr in sc_addrs {
+        let mut rand_seed: [u8; 1072] = [0; 1072];
```
elichai: Why do you generate a whole MegaByte of random just to take the first 32 bytes?

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+        }
+    }
+    Ok(response_data)
+}
+
+/// Get encrypted state keys
+pub(crate) fn ecall_get_enc_state_keys_internal(msg_bytes: Vec<u8>, addrs_bytes: Vec<u8>, sig: [u8; 65], sig_out: &mut [u8; 65]) -> Result<Vec<u8>, EnclaveError> {
+    let msg = PrincipalMessage::from_message(&msg_bytes)?;
+    let sc_addrs: Option<Vec<ContractAddress>> = match msg.data.clone() {
```
elichai: Why an option of vec and not just a vec and then you could return an empty one?

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        match guard.insert(addr.clone(), doc.data) {
+            Some(prev) => println!("New key stored successfully, previous key: {:?}", prev),
+            None => println!("Initial key stored successfully"),
+        }
+        results.push(doc.data);
+    }
+    Ok(results)
+}
+
+fn build_get_state_keys_response(_sc_addrs: Option<Vec<ContractAddress>>) -> Result<Vec<(ContractAddress, StateKey)>, EnclaveError> {
```
elichai: you can receive here a vec and just check if it's empty. no need for the Option.
this will remove an `unwrap` and make the code look nicer.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            });
+        }
+    };
+    let recovered = KeyPair::recover(&msg_bytes, sig)?;
+    println!("Recovered signer address from the message signature: {:?}", recovered.address());
+    // TODO: Verify that the worker is selected for all addresses or throw
+    let response_data = build_get_state_keys_response(sc_addrs)?;
+
+    // Generate the encryption key material
+    let mut rand_num: [u8; 1072] = [0; 1072];
```
elichai: Again, a whole MB for just 44 bytes?

enigma-principal/enclave/src/keys_keeper_t/keeper_types_t.rs
```diff
@@ -0,0 +1,108 @@
...
+        } {}
+        println!("The selected workers: {:?}", selected_workers);
+        Ok(selected_workers)
+    }
+}
+
+impl Decodable for InputWorkerParams {
+    fn decode(rlp: &UntrustedRlp) -> Result<Self, DecoderError> {
+        Ok(Self {
+            block_number: U256::from_bigint(rlp.val_at(0)?),
```
elichai: What's up with these bigints?
I can't see them in the docs of etccommon-rlp, where do they come from? do we really need to have bigint as a dependency? https://docs.rs/etcommon-rlp/0.2.4/rlp/trait.Decodable.html

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    let previous_epoch = get_epoch(&guard, None)?;
+    let nonce: Uint = match previous_epoch {
+        Some(_) => guard.keys().max().unwrap() + 1,
+        None => Uint::from(INIT_NONCE),
+    };
+    println!("Generated a nonce by incrementing the previous by 1 {:?}", nonce);
+    let nonce_bytes: EpochNonce = nonce.into();
+    nonce_out.copy_from_slice(&nonce_bytes[..]);
+
+    // TODO: Check if needs to check the random is within the curve.
```
elichai: curve?

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    rsgx_read_rand(&mut rand_out[..])?;
+    let seed_token = Token::Uint(rand_out[..].into());
+    let seed = seed_token.to_uint().unwrap();
+    println!("Generated random seed: {:?}", seed);
+
+    let epoch = new_epoch(&mut guard, &worker_params, &nonce, &seed)?;
+    let msg = epoch.raw_encode()?;
+    let hash = msg.keccak256();
+    println!("Signing msg hash {} with signer address {}", hash.to_hex(), SIGNING_KEY.get_pubkey().address_string());
+    let sig = SIGNING_KEY.sign(hash.as_ref())?;
```
elichai: You know that the sign message already hash the data. right?

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+    let nonce = get_max_nonce(&guard);
+    let epoch: Epoch = guard.get(&nonce).unwrap().clone();
+    Ok(Some(epoch))
+}
+
+/// Creates new epoch both in the cache and as sealed documents
+fn new_epoch(guard: &mut SgxMutexGuard<HashMap<Uint, Epoch, RandomState>>, worker_params: &InputWorkerParams, nonce: &Uint, seed: &Uint) -> Result<Epoch, EnclaveError> {
+    let mut marker_doc: SealedDocumentStorage<EpochNonce> = SealedDocumentStorage {
+        version: 0x1234, //TODO: what's this?
```
elichai: Right. what is this? :)

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    path_buf.push(EPOCH_DIR);
+    path_buf
+}
+
+fn get_epoch_nonce_path() -> path::PathBuf {
+    get_epoch_root_path().join("nonce.sealed")
+}
+
+fn get_max_nonce(guard: &SgxMutexGuard<HashMap<Uint, Epoch, RandomState>>) -> Uint {
+    guard.keys().max().unwrap().clone()
```
elichai: Can the HashMap be empty?
If so maybe do `unwrap_or_deafult()` to return `0` in the case it's empty?

enigma-principal/app/src/logging.rs
```diff
@@ -0,0 +1,38 @@
+use std::{fs::{self, File}, path::Path};
```
elichai: maybe this file can be moved to `enigma-tools-u`

enigma-principal/app/src/keys_u/km_reader.rs
```diff
@@ -0,0 +1,107 @@
...
+use enigma_crypto::KeyPair;
+use enigma_types::ContractAddress;
+
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+pub enum PrincipalMessageType {
+    Request(Option<Vec<ContractAddress>>),
+}
+
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+pub struct PrincipalMessage {
```
elichai: Why are you deserializing this outside of the enclave?
Why not pass it directly to the enclave and deserialize it && verify the signature there?

enigma-principal/app/src/epoch_u/epoch_types.rs
```diff
@@ -0,0 +1,140 @@
...
+impl IntoBigint<bigint::H256> for H256 { fn bigint(self) -> bigint::H256 { bigint::H256(self.0) } }
+
+impl IntoBigint<bigint::U256> for U256 { fn bigint(self) -> bigint::U256 { bigint::U256(self.0) } }
+
+
+impl Encodable for InputWorkerParams {
+    fn rlp_append(&self, s: &mut RlpStream) {
+        s.begin_list(3);
+        s.append(&self.block_number.bigint());
+        s.append_list(&self.workers.iter().map(|a| a.bigint()).collect::<Vec<bigint::H160>>());
```
elichai: Again, I can't seem to find where the bigint comes from.
https://docs.rs/etcommon-rlp/0.2.4/rlp/trait.Encodable.html

elichai (changes requested):
Ok, this is my first pass of the code.
It's really really hard to review such a big PR. 
next time please try to split into little PRs,
this will help you with the need to always merge from develop and will help me in reviewing.

enigma-tools-u/src/web3_utils/w3utils.rs
```diff
@@ -74,6 +74,11 @@ pub fn connect(url: &str) -> Result<(web3::transports::EventLoopHandle, Web3<Htt
     Ok((_eloop, w3))
 }
 
+pub fn connect_batch(http: Http) -> Web3<Batch<Http>> {
+    let w3 = web3::Web3::new(web3::transports::Batch::new(http));
```
elichai:
```suggestion
    web3::Web3::new(web3::transports::Batch::new(http))
```
You can return this directly

enigma-principal/app/src/esgx/keys_keeper_u.rs
```diff
@@ -0,0 +1,118 @@
...
+            &request.sig.into(),
+            enc_response_slice.as_mut_ptr() as *mut u8,
+            &mut enc_response_len_out,
+            &mut sig_out,
+        )
+    };
+    if retval != EnclaveReturn::Success || status != sgx_status_t::SGX_SUCCESS {
+        return Err(EnclaveFailError { err: retval, status }.into());
+    }
+    let enc_response_out: Vec<u8> = enc_response_slice[0..enc_response_len_out].iter().cloned().collect();
```
elichai:
```suggestion
    let enc_response_out: Vec<u8> = enc_response_slice[0..enc_response_len_out].to_vec()
```
This should be faster and is more readable

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -0,0 +1,267 @@
...
+        let worker_params: InputWorkerParams = InputWorkerParams {
+            block_number,
+            workers: result.0,
+            stakes: result.1,
+        };
+        println!("The active workers: {:?}", worker_params);
+        let epoch_state = &mut set_worker_params(self.eid.load(Ordering::SeqCst), worker_params.clone())?;
+        println!("Waiting for setWorkerParams({:?}, {:?}, {:?})", block_number, epoch_state.seed, epoch_state.sig);
+        // TODO: Consider a retry mechanism, either store the EpochSeed or add a getter ecall
+        let receipt = self.contract.set_workers_params(block_number, epoch_state.seed.clone(), epoch_state.sig.clone(), gas_limit, confirmations)?;
```
elichai: No need to clone the seed, it's a `Copy` type

enigma-principal/app/src/keys_u/keeper_types_u.rs
```diff
@@ -0,0 +1,118 @@
...
+        let mut nonce = U256::from(0);
+        let mut selected_workers: Vec<H160> = Vec::new();
+        while {
+            let token = WorkerSelectionToken { seed, sc_addr, nonce };
+            // This is equivalent to encodePacked in Solidity
+            let hash: [u8; 32] = token.raw_encode()?.keccak256().into();
+            let mut rand_val: U256 = U256::from(hash) % balance_sum;
+            println!("The initial random value: {:?}", rand_val);
+            let mut selected_worker = self.workers[self.workers.len() - 1].clone();
+            for i in 0..self.workers.len() {
```
elichai: why not just `for worker in self.workers` ?

eng-wasm/dispatch/src/lib.rs
fredfortier: I hadn't noticed this. I will investigate.

enigma-principal/app/src/cli/app.rs
fredfortier: This is important for integration testing. The contract deployment script reads the address from here.

enigma-tools-t/src/document_storage_t.rs
fredfortier: Most this file comes from the original `storage_t`. I haven't look at optimizations beyond variable length and error handling. I will do some research on this safer `unseal_data`.

enigma-tools-t/src/document_storage_t.rs
fredfortier: That's just the default value that was already. This should be enough for the data that I'm storing but I'll research this to see if can be more dynamic.

enigma-tools-t/src/document_storage_t.rs
elichai: and it works? I'm actually pretty surprised, because the old `storage_t` was written before we really understood a lot and it's pretty hacky, so I'm surprised it works for bigger data

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: This is how I've need in done in the existing code. I'll research further to better understand the `rand` function.

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
fredfortier: This comes from the original code. I believe that Isan wrote this comment. I don't understand it  either but did not want to lose track of it.

enigma-principal/enclave/src/keys_keeper_t/keeper_types_t.rs
fredfortier: It's required to use this particular ectcommon RLP Crate. These crates use `bigint` instead of `ethereum-types`. I suppose that I could use a different encoding method here since RLP is no longer imperative.

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
fredfortier: Ah! Thanks for reminding me. This explain the issue. I'm not sure how I got it mixed-up.

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
fredfortier: I have no idea. This comes from the original `storage_t`. I tried not to break things that I did not understand. Isan might know.

enigma-principal/app/src/keys_u/km_reader.rs
fredfortier: Here is the logic:
1. The list of contract => selected worker is stored on the untrusted side
2. Here, I read the message to verify if the list of addresses is None. If so, I find the contract addresses for this worker in the epoch and pass to the enclave along with the message.
3. If the message does contain a list of addresses, I just pass the message to the enclave.
4. In the enclave, I deserialize the message again and read the addresses either from the message or, in None, from the list passed by the untrusted. I also verify the signature and run the worker selection algorithm to confirm that the addresses are valid.

This approach achieves two things:
 1. Allow messages to pass-through the the enclave so that the untrusted side cannot censor the addresses that it may contain.
2. Keeps the list of secret contracts on the untrusted side. It would be hard to prove this list to the enclave. At most, the untrusted side could censor secret contract addresses selected for the worker. However, the worker could simply follow-up with a message requesting the missing addresses which the untrusted side cannot modify.

enigma-principal/app/src/epoch_u/epoch_types.rs
fredfortier: Sam answer as above. It's the types primitive for etccommon modules.

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: I wasn't able to pass a u64 across threads but will try again.

enigma-tools-u/src/web3_utils/w3utils.rs
fredfortier: You're right. I'm not using the batch transport anymore. I will clean-up.

enigma-tools-t/src/document_storage_t.rs
fredfortier: Yes, it works for variable size of byte arrays. I'm not trying to store large documents at the moment. Do you have an example of a more mature implementation of storage which I can review?

enigma-principal/app/src/keys_u/keeper_types_u.rs
fredfortier: Because I need the index in the line below.

eng-wasm/dispatch/src/lib.rs
fredfortier: Fixed

enigma-core/app/enigma-core-app.iml
fredfortier: Fixed

enigma-principal/app/src/boot_network/keys_provider_http.rs
fredfortier: Good catch! Fixed in commit a2c322bc89442eaa0f2b175f400fcf53eb424988.

enigma-principal/app/src/boot_network/keys_provider_http.rs
fredfortier: The parameter was removing CORS header verification. I set it for troubleshooting as one point in time but you are correct that it's not needed. Removed in commit 5d2285cdfe8d20f05d7b9c88a11a173efc93ad4f.

enigma-principal/app/src/boot_network/principal_manager.rs
fredfortier: Interesting, I did not realize that implementing the `Copy` trait implicitly cloned the variable instead of moving. I thought that it still needed to be specified explicitly. My IDE does not seem to support this but the compiler is happy with it. I'll inspect throughout and create a single commit.

enigma-principal/app/src/boot_network/principal_manager.rs
fredfortier: Yes, I forgot to handle this. Fixed in commit 031fd8c363fa5688e64bd95558f29540aca50e14.

enigma-principal/enclave/src/lib.rs
fredfortier: Fixed in commit e7adf35481143153d4b637aa8f6ae93499172d2e.

enigma-principal/app/src/boot_network/principal_manager.rs
fredfortier: Fixed in commit: e7adf35481143153d4b637aa8f6ae93499172d2e.

enigma-principal/app/src/cli/app.rs
fredfortier: Fixed in 68afcef246bbe4978d4ab94501853b45ead7fd67.

enigma-tools-t/src/document_storage_t.rs
fredfortier: This is just an adaptation of `storage_t` which I assumed was current. I need to better understand the storage service to answer this. Is there a better example than `storage_t` in our code? If not, I research in the Baidu SDK.

enigma-tools-t/Cargo.toml
fredfortier: Forgot to remove the feature after experimenting with it. Fixed in commit bed425e47186a9933bc4329680ffa9aeb31ad2cb.

enigma-tools-t/src/document_storage_t.rs
fredfortier: Removed in commit b9f50bc39689924bc12c3f0e14ee5792b72c861d.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: I implemented it like this for a specific reason, putting the `key.is_none()` logic in the match fn gives an immutable (`guard get`) and mutable (`guard.insert`) borrow conflict.

enigma-tools-u/src/web3_utils/w3utils.rs
fredfortier: Removed `connect_batch` in commit b9f50bc39689924bc12c3f0e14ee5792b72c861d.

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Adjusted in commit 762ae1c73e8e00ab8c54fcc6c4e9c5eb788d1d32.

enigma-principal/app/src/logging.rs
fredfortier: Refactored in commit f548eafbf818bd9c44486a3f4b0d673971e97afc. There is more code copy/pasted especially in `enigma-principal/app/src/esgx` module. I tried to not to question functioning code and limit impact on shared modules, but it's probably worth refactoring this area.

enigma-principal/app/src/boot_network/principal_manager.rs
fredfortier: Yes, I explained this in the notes included in the PR. That said, I agree that this test is useful. I just did not prioritize refactoring it since I have another way to test. I will re-enable it shortly.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Fixed in commit 1df39b26a9e42561bd03defb2c5136408846a391.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Thanks for catching this. Fixed in commit 1df39b26a9e42561bd03defb2c5136408846a391.

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
fredfortier: No, it can't, there is a guard clause before its usage. Since this fn was only used once, I refactored it to put the logic directly after the guard clause. I can see how this wasn't very readable. Commit dd39e025a5b34eecd022a5071f6588798cdc49cd.

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
fredfortier: Removed the TODO in 3d1f5d0f650db92fb9f3f0ff3125074c497d4cc6.

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
fredfortier: Removed the redundant hashing in commit 3d1f5d0f650db92fb9f3f0ff3125074c497d4cc6.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: You're right, the option is unnecessary and the code looks cleaner without. Commit e6592f92c553e998e5dbb454861702eb027b2d61.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Refactored in commit e6592f92c553e998e5dbb454861702eb027b2d61.

enigma-tools-t/src/document_storage_t.rs
fredfortier: Obviously :smile: Commit 1bcbdd7da8841587d357af9b77be98dea69b83eb.

enigma-principal/app/src/keys_u/keeper_types_u.rs
elichai: Ok, FYI best practice is `enumerate` https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate

enigma-principal/enclave/src/epoch_keeper_t/mod.rs
elichai: I think this is a feature that @Isan-Rivkin thought might be useful for the signing keys, I don't think there's a use case here and even in the signing keys we didn't end up using this.

dockerfile/run.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
elichai, line 1: I'm not sure if that's the right place for this. e.g. I just had this in my `.bash_aliases`/`.bashrc`:
```
alias enigma-docker='docker run --security-opt seccomp:unconfined --cap-add=SYS_PTRACE --net="host" -v /home/elichai2/CLionProjects/enigma-core:/home/elichai2/enigma-core -v /home/elichai2/.cargo-enigma-docker:/home/elichai2/.cargo -ti --device /dev/isgx enigma_dev:latest'
```
So I don't think that your exact preferences should be in the git

.gitignore
```diff
@@ -85,3 +86,9 @@ fabric.properties
 sgx-tests
 CMakeLists.txt
 dockerfile-elichai/
+
+/target
+**/*.rs.bk
+
+/target
+**/*.rs.bk
```
elichai: You wrote this twice.

enigma-principal/enclave/src/lib.rs
```diff
@@ -59,50 +74,63 @@ fn get_sealed_keys_wrapper() -> asymmetric::KeyPair {
 
 #[no_mangle]
 pub extern "C" fn ecall_get_signing_address(pubkey: &mut [u8; 42]) {
+    println!("The priv key for worker 8: {:?}", SIGNING_KEY.get_privkey().to_vec().to_hex());
```
elichai: I'm gessing you meant to remove this.

enigma-principal/enclave/src/keys_keeper_t/keeper_types_t.rs
elichai: hmm ok, I still haven't looked into where is this used but I also don't like the endianess assumeption here.
what if `ethereum-types::U256` and `etcommon-bigint::U256` don't use the same endianess?

enigma-principal/enclave/src/lib.rs
fredfortier: Fixed c8ecac858baab8e1445c524e6103b2c6df3d3f76

.gitignore
fredfortier: Fixed 90deb885612469ce15129c979a16cabe78c09549

enigma-principal/app/src/boot_network/principal_manager.rs
fredfortier: Fixed a62f7c406935853558959cd57ceed2300154e3ce

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Fixed a62f7c406935853558959cd57ceed2300154e3ce

enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+impl TryInto<[u8; 65]> for StringWrapper {
+    type Error = Error;
+
+    fn try_into(self) -> Result<[u8; 65], Self::Error> {
+        let value = &self;
+        let bytes = value.0.from_hex()?;
+        let mut slice: [u8; 65] = [0; 65];
+        slice.copy_from_slice(&bytes[..]);
```
moriaab: The length of source must be the same as self in copy_from_slice, the function will panic otherwise. String can be any length. You should check it somehow or do it safer.

enigma-principal/app/src/main.rs
```diff
@@ -1,34 +1,61 @@
...
 mod cli;
 mod common_u;
 mod esgx;
-
-pub use esgx::general::ocall_get_home;
+mod epoch_u;
+mod keys_u;
 
 #[allow(unused_variables, unused_mut)]
 fn main() {
```
moriaab, line 43: @elichai, @fredfortier
This is a general comment. We have duplicated code in functions `main` in enigma-core/app/src/main.rs and enigma-principal/app/src/main.rs: from the first line to getting eid. I think it is better to extract it to untrusted common utils.

enigma-principal/app/src/esgx/equote.rs
```diff
@@ -1,52 +1,57 @@
...
+
 use failure::Error;
 use sgx_types::*;
-use std::*;
+
+use common_u::errors;
 
 extern "C" {
-    fn ecall_get_signing_address(eid: sgx_enclave_id_t, pubkey: &mut [u8; 42]) -> sgx_status_t;
+    fn ecall_get_signing_address(eid: sgx_enclave_id_t, pubkey: &mut [u8; 20]) -> sgx_status_t;
```
elichai, line 7: You changed the signature here but didn't change it in the EDL or in the trusted side.
This means the edgerator just memcpy's another 22 bytes that just exist after this array. I'm suprised you didn't have any segmentation faults.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    let response_msg = PrincipalMessage::new_id(response_msg_data, id, pubkey);
+    if !response_msg.is_response() {
+        return Err(EnclaveError::KeyProvisionError {
+            err: "Unable create response".to_string()
+        });
+    }
+    // Generate the iv from the first 12 bytes of a new random number
+    let mut iv: [u8; 12] = [0; 12];
+    iv.clone_from_slice(&rand_num[32..44]);
+    let response = response_msg.encrypt_with_nonce(&derived_key, Some(iv))?;
```
elichai: `encrypt_with_nonce` shouldn't be called directly, it's for internal use and tests.
I'll make a PR that adds a warning there.

enigma-principal/enclave/src/lib.rs
```diff
@@ -62,47 +77,59 @@ pub extern "C" fn ecall_get_signing_address(pubkey: &mut [u8; 42]) {
...
+    EnclaveReturn::Success
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn ecall_get_enc_state_keys(msg: *const u8, msg_len: usize,
+                                                  addrs: *const u8, addrs_len: usize, sig: &[u8; 65],
+                                                  enc_response_out: *mut u8, enc_response_len: &mut usize,
+                                                  sig_out: &mut [u8; 65]) -> EnclaveReturn {
+    let msg_bytes = slice::from_raw_parts(msg, msg_len).to_vec();
+    let addrs_bytes = slice::from_raw_parts(addrs, addrs_len).to_vec();
```
elichai: No need to convert these to vectors, you can pass them as slices directly.

enigma-principal/app/src/esgx/keys_keeper_u.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+        None => vec![0],
+    };
+    let status = unsafe {
+        ecall_get_enc_state_keys(
+            eid,
+            &mut retval,
+            msg_bytes.as_ptr() as *const u8,
+            msg_bytes.len(),
+            addrs_bytes.as_ptr() as *const u8,
```
elichai: We should always use `as_c_ptr()`(https://github.com/enigmampc/enigma-core-internal/blob/develop/enigma-types/src/traits.rs#L3) otherwise if the slice/vec is empty it will return a ptr 0x0, which isn't a defined behavior in C, so this causes SegFaults.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    println!("Recovered signer address from the message signature: {:?}", recovered.address());
+    // TODO: Verify that the worker is selected for all addresses or throw
+    let response_data = build_get_state_keys_response(sc_addrs)?;
+
+    // Generate the encryption key material
+    let mut rand_num: [u8; 44] = [0; 44];
+    rsgx_read_rand(&mut rand_num)?;
+    let mut privkey_slice = [0u8; 32];
+    privkey_slice.copy_from_slice(&rand_num[..32]);
+    let key_pair = KeyPair::from_slice(&privkey_slice)?;
```
elichai: Please use `KeyPair::new()` as the generated private key might not be valid for the curve. it should be checked (which `KeyPair::new()` does)

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    let derived_key = key_pair.derive_key(&msg.get_pubkey())?;
+
+    // Create the response message
+    let response_msg_data = PrincipalMessageType::Response(response_data);
+    let id = msg.get_id();
+    let pubkey = key_pair.get_pubkey();
+    let response_msg = PrincipalMessage::new_id(response_msg_data, id, pubkey);
+    if !response_msg.is_response() {
```
elichai: Why do you check if it's not a response right after you made it as a response?

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+    // Generate the iv from the first 12 bytes of a new random number
+    let mut iv: [u8; 12] = [0; 12];
+    iv.clone_from_slice(&rand_num[32..44]);
+    let response = response_msg.encrypt_with_nonce(&derived_key, Some(iv))?;
+    let response_bytes = response.to_message()?;
+    println!("The partially encrypted response: {:?}", response_bytes.to_hex());
+    // Signing the encrypted response
+    // This is important because the response might be delivered by an intermediary
+    let hash = response_bytes.clone().keccak256();
```
elichai: You're hashing before the sign function which already hashes for you.

enigma-principal/app/src/boot_network/principal_manager.rs
fredfortier: Fixed the unit test (minus contract deployment) 31888b32e95a4a7471e809442e465a89c9195c36

enigma-principal/app/src/esgx/equote.rs
fredfortier: The trusted side looks fine but I fixed the mismatching EDL f58c88976c52619e382b724ddb31587d204d5dad

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: I don't see any abstraction in the `km_primitive` module.

enigma-principal/enclave/src/lib.rs
fredfortier: My compiler wants me to. I'm not sure why it's using a `Vec` implementation. I believe that I'm importing it correctly.
```
error[E0308]: mismatched types
   --> src/lib.rs:103:60
    |
103 |     let response = match ecall_get_enc_state_keys_internal(*msg_bytes, *addrs_bytes, *sig, sig_out) {
    |                                                            ^^^^^^^^^^
    |                                                            |
    |                                                            expected struct `serde::export::Vec`, found slice
    |                                                            help: try using a conversion method: `*msg_bytes.to_vec()`
    |
    = note: expected type `serde::export::Vec<u8>`
               found type `[u8]`
```

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Fixed 951089c26472ff38de1975e13f5c7bb492e6db27

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: I was trying to account for the possibility that the message is not initialized correctly. After looking more closely at what `new_id` does, I agree that it is redundant. Fixed 951089c26472ff38de1975e13f5c7bb492e6db27

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Fixed 951089c26472ff38de1975e13f5c7bb492e6db27

enigma-principal/enclave/src/keys_keeper_t/mod.rs
elichai: I wrote that trait in a way that you need to implement `encrypt_with_nonce` and you get `encrypt` for free https://github.com/enigmampc/enigma-core-internal/blob/develop/enigma-crypto/src/lib.rs#L37

enigma-principal/enclave/src/lib.rs
elichai: Change `ecall_get_enc_state_keys_internal` to accept slices

enigma-principal/app/src/esgx/equote.rs
elichai: That's because you fixed it here :) https://github.com/enigmampc/enigma-core-internal/pull/89/commits/45a294ae75c62b62b91020fd053521727a4f7545

elichai (approved):

PR #94, opened 2019-03-17
Upgrading to Baidu SDK 1.0.6

elichai:
This PR upgrades the SDK to 1.0.6 which also upgrades the rustc version to `nightly-2019-01-28`

This can wait for after the #89 Is merged and even after the release if we prefer


elichai:
This has been superseded by #157

PR #101, opened 2019-03-25
formattings and remove unnecessary imports in km-node 

elichai:
- removes a lot of unused dependencies in `enigma-principal` which should make the build faster.
- removed unnecessary imports and combined imports together.


elichai:
After this I feel comfortable merging `principal-keymgmt` into develop.

fredfortier (approved):
Looks good

PR #104, opened 2019-03-26
Coin toss contract

moriaab:


enigma-runtime-t/src/data/state.rs
```diff
@@ -46,6 +46,12 @@ impl IOInterface<EnclaveError, u8> for ContractState {
         self.json[key] = value.clone();
         Ok(())
     }
+
+    fn remove_key(&mut self, key: &str) {
+        if let Some(ref mut v) = self.json.as_object_mut() {
```
elichai, line 51: are we sure it's always an object?

elichai (approved):
LGTM

enigma-runtime-t/src/data/state.rs
moriaab: For now we are sure, it is created as follows: let json = serde_json::from_str("{}").unwrap();

PR #106, opened 2019-03-27
Updated Rocksdb

elichai:
So after my latest PR was merged https://github.com/rust-rocksdb/rust-rocksdb/pull/284 and they released a new version (0.12.1) we can stop using my fork and move to using the upstream version from crates.io.

This also means upgrading facebooks rocksdb C++ from 5.14.2 to 5.17.2

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,7 +1,7 @@
 pipeline:
   core:
     group: tests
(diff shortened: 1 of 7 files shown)


AvishaiW (approved):
finally :)
looks great!

PR #107, opened 2019-03-28
Millionaires problem demo

apalepu23:


examples/eng_wasm_contracts/millionaires_problem_demo/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct Millionaire {
+    address: H256, // field containing 32 byte hash type for millionaire's address
+    net_worth: U256, // field containing 32 byte uint for millionaire's net worth
+}
+
+// Public-facing secret contract function declarations
+#[pub_interface]
+pub trait ContractInterface{
+    fn construct();
```
AvishaiW: if empty, not needed

examples/eng_wasm_contracts/millionaires_problem_demo/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+            Some(vec) => vec,
+            None => Vec::new(),
+        }
+    }
+}
+
+impl ContractInterface for Contract {
+    // Empty constructor
+    #[no_mangle]
+    fn construct() {
```
AvishaiW: if empty, not needed

moriaab:
@apalepu23 the .iml file should be ignored, please insert to .gitignore.

examples/eng_wasm_contracts/millionaires_problem_demo/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+extern crate eng_wasm_derive;
+extern crate rustc_hex as hex;
+#[macro_use]
+extern crate serde_derive;
+extern crate serde;
+
+use eng_wasm::*;
+use eng_wasm_derive::pub_interface;
+use eng_wasm::String;
+use eng_wasm::from_utf8;
```
moriaab: There is `use eng_wasm::*` in code, therefore `use eng_wasm::String` and 
'use eng_wasm::from_utf8` are redundant.

examples/eng_wasm_contracts/millionaires_problem_demo/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+use hex::ToHex;
```
moriaab: It seems like FromHex and ToHex are not used, therefore lines 17 and 18 may be deleted.

examples/eng_wasm_contracts/millionaires_problem_demo/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+        // Write the updated vector to contract's state
+        write_state!(MILLIONAIRES => millionaires);
+    }
+
+    // Compute the richest millionaire by returning the 32-byte hash type for the address
+    #[no_mangle]
+    fn compute_richest() -> H256 {
+        // Read state to get vector of Millionaires and obtain the struct corresponding to the
+        // richest millionaire by net worth
+        match Self::get_millionaires().iter().max_by_key(|m| m.net_worth) {
```
moriaab, line 67: If there are several equal maximums, only one is returned. This is Ok, if this was your intention. But maybe it is better to return a vector of addresses? I am not sure though.

examples/eng_wasm_contracts/millionaires_problem_demo/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
moriaab: What is it for?

moriaab:
@AvishaiW suggested that you run rust-clippy on your code (https://github.com/rust-lang/rust-clippy#usage), which is actually a very good practice.

moriaab (approved):

examples/eng_wasm_contracts/millionaires_problem_demo/src/main.rs
apalepu23: removed!

examples/eng_wasm_contracts/millionaires_problem_demo/src/lib.rs
apalepu23: i felt returning multiple in this scenario gives you information about the value of another person's net worth (in the case it is equal), which goes against the motivation of the millionaires' problem. It may be fine given this edge case, but that was my reasoning behind this, but don't personally care too much either way! Leaving it as is for now :)

PR #108, opened 2019-04-04
Delta previous hash was the hash of the unencrypted delta

elichai:
This PR contains 2 important fixes:
1. `Delta.privous_hash` and `state.delta_hash` contained the hash of the unencrypted delta, now it contains the hash of the encrypted delta.
2. because an empty state has `delta_index=0` we couldn't distinguish between an empty state and a state after the first delta, meaning it always skipped delta `0`,
now it checks if `state.delta_hash == [0u8; 32]` to determine if it has a delta already or if it's an empty state.

What I **don't** like about this:
1. `into_result` now need to get the state key.
2. `apply_delta` now applies and decrypts the delta, which violates the single responsibility principle and makes the it harder to test (as you can see in the tests I had to change)
3. right now `get_state_key` copies the stateKey in memory, a practice we should try to minimize and in the future maybe avoid altogether, but that's not a problem for now. (we'll want to zero out the memory were a secret was staying)


enigma-runtime-t/src/data/state.rs
```diff
@@ -54,28 +55,34 @@ impl IOInterface<EnclaveError, u8> for ContractState {
...
+        self.delta_hash = delta_hash;
+        self.delta_index = dec_delta.index;
         Ok(())
     }
 
-    fn generate_delta_and_update_state(old: &Self, new: &mut Self) -> Result<StatePatch, EnclaveError> {
-        new.delta_index = &old.delta_index+1;
-        let result = StatePatch{
+    fn generate_delta_and_update_state(old: &Self, new: &mut Self, key: &'a StateKey) -> Result<EncryptedPatch, EnclaveError> {
+        if old.delta_hash.is_zero() {
```
moriaab, line 72: Maybe it is better to extract the code of getting delta index from delta to a function, this will be cleaner. 
I think that the differentiation of the empty and non-empty delta this way is fine.

enigma-core/app/src/km_u.rs
```diff
@@ -200,128 +202,138 @@ pub mod tests {
     fn test_the_whole_round() {
         let (mut db, _dir) = create_test_db();
         //Making a request
-        let address = fill_the_db(&mut db);
+        let (addresses, keys) = fill_the_db(&mut db);
+        let keys = keys.into_iter()
+            .zip_longest(addresses.iter())
```
moriaab, line 207: Rust loves iterators :-)

moriaab:
As we discussed we will do the refactoring of the Runtime. All the functions related to execution (including work with deltas and states) will be done in Runtime and what is needed outside the Runtime will be  returned as RuntimeResult. After into_result the Runtime will be destroyed. This design should eliminate  part of the concerns you raised.
The concern of hardness of tests cannot be something that affect the design.

moriaab:
This is very impressive how you did this cross-project fix so fast!

moriaab (approved):

Issue #110, opened 2019-04-05
KeyMgmt node crashed when setting workerParams when there is a contract deployed


lacabra:
**Describe the bug**
The Key Management node crashes when trying to set Worker Parameters *AND* there is at least one secret contract deployed in the network. When there are no contracts deployed, it works just fine.

**To Reproduce**
1. Launch the network
2. Deploy a secret contract
3. Trigger an epoch change

**Expected behavior**
The Key Management node should handle epoch changes with secret contracts deployed.

**Backtrace**
```
principal_1   | [? ] Blocks @ previous: 44, current: 54, next: 54 [? ]
principal_1   | [? ] New epoch found [? ]
principal_1   | The active workers: InputWorkerParams { block_number: 54, workers: [0xe43e4e5aeae608861b0d8d7c9533d22204b08181], stakes: [90000000000] }
principal_1   | Successfully decoded RLP worker parameters
principal_1   | Getting epoch for block number: None
principal_1   | Generated a nonce by incrementing the previous by 1 0x2
principal_1   | Generated random seed: 0xda927dd3be310c65d528cf260380ee33eeb95dbaf34516dd16a85f81ada03ce6
principal_1   | [+] Home dir is /root
principal_1   | Back from Ocall: /root/.enigma
principal_1   | Sealed document: "/root/.enigma/epoch/nonce.sealed" written successfully.
principal_1   | Sealed the epoch marker: "/root/.enigma/epoch/nonce.sealed"
principal_1   | Storing epoch: Epoch { nonce: 0x2, seed: 0xda927dd3be310c65d528cf260380ee33eeb95dbaf34516dd16a85f81ada03ce6, worker_params: InputWorkerParams { block_number: 0x36, workers: [0xe43e4e5aeae608861b0d8d7c9533d22204b08181], stakes: [0x14f46b0400] } }
principal_1   | Initial epoch stored successfully
principal_1   | Signed the message : 0xda927dd3be310c65d528cf260380ee33eeb95dbaf34516dd16a85f81ada03ce60000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000001000000000000000000000000e43e4e5aeae608861b0d8d7c9533d22204b08181000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000014f46b0400
[...]

PR #112, opened 2019-04-05
FIX: Fixed an issue with using Hash256 in the epoch state and added a…

fredfortier:
It seems like Serde does not automatically serialize Struct HashMap keys so I implemented a custom serializer. I also added a unit test to catch this sort of issue in the future. This was caught be `Enigma.spec.js` but unit tests did not have sufficient coverage.

This should resolve issue #110.


lacabra:
I confirm that this PR does indeed resolve the issue in the docker network. Thank you

elichai:
the problem is that you use serde-json to serialize the struct and JSON's keys are required to be strings.
I think the right solution is to just not use serde-json and instead use a different serializer (i.e. MessagePack)

it will work:
```
use rmp_serde::{Serializer, Deserializer};
use serde::{Serialize, Deserialize};
use std::collections::HashMap;


#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
pub struct Hash256(pub [u8; 32]);

#[derive(Serialize, Deserialize, PartialEq, Debug)]
struct Test {
    pub map: HashMap<Hash256, u32>,
}


fn main() {
    let mut a = Test{map: HashMap::new()};
    a.map.insert(Hash256([17u8; 32]), 58);
    let mut buf = Vec::new();
    a.serialize(&mut Serializer::new(&mut buf)).unwrap();
    let mut des = Deserializer::new(&buf[..]);
     let b: Test = Deserialize::deserialize(&mut des).unwrap();
     assert_eq!(a,b);
}
```

fredfortier:
I don't see how using MessagePack over JSON is the "right solution" but I'll make the suggested change.

elichai:
Because what this is doing is lying that hash256 is actually hex. And converting it back and forth.
When the obvious solution is that the serialization you used wasn't suitable.

Correct me if I'm wrong but no human should understand/read this, Right?

fredfortier:
All other things being equal, I would prefer serializing the epoch state in a human readable format if only to help investigate a failure scenario. That said, it is a marginal benefit and I'm not opposed to your suggestion. I just do not believe that phrasing your suggestion in the absolute form is appropriate.

fredfortier:
Implemented MessagePack encoding.

elichai (approved):
Looks good.
I do encourage you to use the typedefs they are there for a reason (both ethereum-types Address, and our ContractAddress)

fredfortier:
I was unable to push the merged changes to `develop`.

```
[fredfortier@localhost enigma-core-internal]$ git push
Enumerating objects: 13, done.
Counting objects: 100% (13/13), done.
Delta compression using up to 8 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 1.26 KiB | 643.00 KiB/s, done.
Total 5 (delta 4), reused 0 (delta 0)
remote: Resolving deltas: 100% (4/4), completed with 4 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/develop.
remote: error: Required status check "continuous-integration/drone/push" is expected.
To github.com:enigmampc/enigma-core.git
 ! [remote rejected] develop -> develop (protected branch hook declined)
error: failed to push some refs to 'git@github.com:enigmampc/enigma-core.git'
```
I'm not sure to understand this error. Please give me some pointers or feel free to merge on your end.

PR #116, opened 2019-04-10
modified the signing in the erc20 contract from regular signing to multiple signing

AvishaiW:
…to multiple signing.


elichai:
I think we should discuss removing all `usize` instances in the code, and cast to u64.
@moriaab

elichai (approved):
LGTM

PR #120, opened 2019-04-16
Testing a fix for web3 and rustc-mangle

elichai:
This should fix #119 and and change web3 to be retrieved from crates.io

PR #121, opened 2019-04-17
Worker selection guard + Additional Unit Tests

fredfortier:
Key features added:
- More unit tests
- Epoch state storage: The program can resume after shutdown and continue serving the active epoch
- Worker selection guard: Only selected workers are allowed to request keys for the active epoch


enigma-core/app/Cargo.toml
```diff
@@ -25,7 +25,7 @@ log = "0.4.6"
 log-derive = "0.2"
 simplelog = "0.5.3"
 structopt = "0.2"
-
+rustc-demangle = "<0.1.14"
```
elichai: Can you please rebase this *without* the stuff on `rustc-demangle` and `rust-web3` that were already fixed https://github.com/enigmampc/enigma-core/pull/120

enigma-principal/app/src/cli/app.rs
```diff
@@ -70,7 +70,11 @@ pub fn start(eid: sgx_enclave_id_t) -> Result<(), Error> {
             let block_number = principal.get_block_number()?;
             let tx = epoch_provider.set_worker_params(block_number, gas_limit, principal_config.confirmations as usize)?;
             println!("The setWorkersParams tx: {:?}", tx);
-        } else if opt.get_state_keys.is_some() {
+        }  else if opt.confirm_worker_params {
```
elichai, line 73: What is this doing?

enigma-principal/app/src/esgx/epoch_keeper_u.rs
```diff
@@ -34,44 +42,59 @@ pub fn set_worker_params(eid: sgx_enclave_id_t, worker_params: &InputWorkerParam
...
+            &seed_in,
+            &nonce_in,
             &mut rand_out,
             &mut nonce_out,
             &mut sig_out,
         )
     };
     if retval != EnclaveReturn::Success || status != sgx_status_t::SGX_SUCCESS {
-        return Err(EnclaveFailError { err: retval, status }.into());
+        bail!("{:?} encountered in the enclave", retval);
```
elichai: Why did you change that?
`bail!` should be used as a temporary/generic error. `EnclaveFailError` is the error that fit here, and we nmeed both the `status` and the `retval` and you just throw away the `status`

enigma-principal/app/src/esgx/keys_keeper_u.rs
```diff
@@ -44,7 +45,7 @@ pub fn get_enc_state_keys(eid: sgx_enclave_id_t, request: StateKeyRequest, epoch
         )
     };
     if retval != EnclaveReturn::Success || status != sgx_status_t::SGX_SUCCESS {
-        return Err(EnclaveFailError { err: retval, status }.into());
+        bail!("{:?} encountered in the enclave", retval);
```
elichai: Same as above

enigma-principal/app/src/main.rs
```diff
@@ -13,6 +13,7 @@ extern crate ethabi;
 #[macro_use]
 extern crate failure;
 extern crate jsonrpc_http_server;
+extern crate jsonrpc_test;
```
elichai: Please move this to `dev-dependencies` and `extern crate` in the test.

enigma-principal/enclave/Enclave.config.xml
```diff
@@ -4,7 +4,7 @@
   <ISVSVN>0</ISVSVN>
   <StackMaxSize>0x40000</StackMaxSize>
   <HeapMaxSize>0x100000</HeapMaxSize>
-  <TCSNum>1</TCSNum>
+  <TCSNum>10</TCSNum>
```
elichai: I'm not sure if this is a good idea.
Do you know what does this do? 
do you know if there's any consequences for this?

enigma-principal/enclave/src/epoch_keeper_t/epoch_t.rs
```diff
@@ -28,12 +30,27 @@ impl Epoch {
...
     /// Encode the Epoch as Ethereum ABI parameters
     fn raw_encode(&self) -> Bytes {
-        let tokens = vec![
-            Token::Uint(self.seed),
-            Token::Uint(self.nonce),
-            Token::Array(self.worker_params.workers.iter().map(|a| Token::Address(*a)).collect()),
-            Token::Array(self.worker_params.stakes.iter().map(|s| Token::Uint(*s)).collect()),
-        ];
-        encode(&tokens)
+        let raw_seed = H256::from(self.seed).0.to_vec();
```
elichai, line 33: Can you please explain what are you doing here exactly? why don't use use the regular `hash_multiple`? Implementing the same thing multiple times is a *bad* idea.
(especially something as sensitive as this)

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -135,16 +138,25 @@ pub(crate) fn ecall_get_enc_state_keys_internal(
     let msg = PrincipalMessage::from_message(msg_bytes)?;
     let user_pubkey = msg.get_pubkey();
     let msg_id = msg.get_id();
-    let sc_addrs: Vec<ContractAddress> = match msg.data {
+    let sc_addrs: Vec<ContractAddress> = match msg.data.clone() {
```
elichai: You can just move the `image` and the `print` to before this and then you won't need to clone anything...

enigma-tools-m/src/keeper_types.rs
```diff
@@ -54,24 +55,30 @@ impl InputWorkerParams {
...
     ///
+    #[logfn(DEBUG)]
     pub fn get_selected_worker(&self, sc_addr: ContractAddress, seed: U256) -> Option<Address> {
+        debug!("Finding selected worker for sc_addr: {:?} and seed: {:?}", sc_addr, seed);
         let workers = self.get_selected_workers(sc_addr, seed, None);
         if workers.is_empty() {
             None
         } else {
-            Some(workers[0])
+            Some(workers[0].clone())
```
elichai: why do you clone this?

elichai (changes requested):
Please please try to make each PR do just one thing.
It's really hard to review big PRs,
every PR should implement one new feature or fix one bug, please.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -216,19 +261,44 @@ impl EpochProvider {
...
     use enigma_crypto::KeyPair;
+    use enigma_tools_u::{esgx::general::storage_dir};
     use enigma_types::ContractAddress;
 
     use super::*;
 
     pub const WORKER_SIGN_ADDRESS: [u8; 20] =
         [95, 53, 26, 193, 96, 206, 55, 206, 15, 120, 191, 101, 13, 44, 28, 237, 80, 151, 54, 182];
 
+    pub fn setup_epoch_storage() {
```
elichai, line 284: What is this complicated thing actually do? why no just create the dir?
and this runs *only* in tests. which means that if the directory isn't existing yet in real code then this won't work.
what do you think about this fix? https://github.com/enigmampc/enigma-core/pull/132

fredfortier: I replied to #132 about why this is only an issue in tests. This "complicated thing" creates the directory and ensures that it is empty (so that one test does not accidentally read the state from another).

enigma-principal/app/src/esgx/keys_keeper_u.rs
fredfortier: You're right. I should have put a TODO here. It was swallowing the exception so I did this temporarily. I'll take a second look and do it properly.

enigma-principal/enclave/Enclave.config.xml
fredfortier: I read about the thread model. I understand that this allows more than one thread to execute in the worker. I also see that it works without noticeable side-effect. As mentioned in the issue description, I'm still investigating this to understand what causes the HTTP server to spawn another thread in the first place. My understanding was that the requests were queued in the Tokio event loop. This confirms it: https://github.com/paritytech/jsonrpc/issues/423. I'm in the process of investigating what causes the second thread. So far, I could only reproduce this in the integration test.

So to answer is that I'm not sure either, this is a solution that works while I continue to investigate.

enigma-principal/enclave/src/epoch_keeper_t/epoch_t.rs
fredfortier: Yes, that's fine. I wasn't aware of the `hash_multiple` then but I have no problem using it. That said, I don't believe that `hash_multiple` behave as expected so the signature won't recover. I just need to better understand how `hash_multiple` represents array lengths.

enigma-principal/enclave/Enclave.config.xml
elichai: Found something here:
https://download.01.org/intel-sgx/linux-2.5/docs/Intel_SGX_Developer_Reference_Linux_2.5_Open_Source.pdf
page 65. 

```
TCSMaxNum, TCSNum, and TCSMinPool are used to determine how many
threads will be created after the enclave initialization, and how many threads
can be created dynamically when the enclave is running. These two kinds of
threads are referred to as static threads and dynamic threads respectively.
```


Do you launch threads?

fredfortier: I read this as well. My understanding is that threads in the enclave are spawned by the untrusted component. I don't spawn threads so I suspected that the JSON-RPC HTTP server did under some circumstances. However, it does not seem to be the case: https://github.com/paritytech/jsonrpc/issues/423 at least in theory.

enigma-core/app/Cargo.toml
fredfortier: Fixed in commit f6d9b00565c15b58edc6820438141cdb9028ee28

enigma-principal/app/src/cli/app.rs
fredfortier: Serving a `getStateKeys` request and signature from the CLI.

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Merged PR #132, moving test setup logic to the main business logic. The `setup` fn is still required to clear the state before each test that touches it.

enigma-principal/app/src/esgx/epoch_keeper_u.rs
fredfortier: Fixed in commit 3f28b0a0cf3d1edf20ec81249b4dcd56bb3b79d0. I'll add some negative test cases to make sure that some trace of the enclave error gets carried over.

enigma-principal/app/src/esgx/keys_keeper_u.rs
fredfortier: Fixed in commit 3f28b0a0cf3d1edf20ec81249b4dcd56bb3b79d0. I'll add some negative test cases to make sure that some trace of the enclave error gets carried over.

enigma-principal/app/src/main.rs
fredfortier: Fixed a88ef5dd2e746f84d0f0df904a0b9aec705ad763

enigma-principal/enclave/Enclave.config.xml
fredfortier: Fixed in commit 1e97be68b6f1124ec768e5e034770dbe798b536e

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Changed in commit 3d17468c6b94cff2820d849c70332b302975c35c. I added a comment to explain why some logic happens before the guard.

enigma-tools-m/src/keeper_types.rs
fredfortier: The borrow checked in my IDE wants me to. Fixed in commit 08fc683d6aefa8bb86ffac395736954ea56b0f4a.

fredfortier (commented):
Addressed each comment and made changes as appropriate.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -223,18 +267,29 @@ impl EpochProvider {
...
+        let mut path = storage_dir(ENCLAVE_DIR).unwrap();
+        path.push(EPOCH_DIR);
+        fs::remove_dir_all(path).unwrap_or_else(|_| println!("Epoch dir already empty"));
```
elichai, line 287: FYI you have `expect(msg)` for this

enigma-principal/app/src/esgx/epoch_keeper_u.rs
```diff
@@ -21,10 +24,16 @@ extern "C" {
...
     let mut retval: EnclaveReturn = EnclaveReturn::Success;
-    let mut nonce_out: [u8; 32] = [0; 32];
-    let mut rand_out: [u8; 32] = [0; 32];
+    println!("Evaluating nonce/seed based on EpochState: {:?}", epoch_state);
+    let (nonce_in, seed_in) = match epoch_state.clone() {
+        Some(e) => (H256::from(e.nonce).0, H256::from(e.seed).0),
+        None => ([0; 32], [0; 32])
+    };
+    println!("Calling enclave set_worker_params with nonce/seed: {:?}/{:?}", nonce_in.to_vec().to_hex(), seed_in.to_vec().to_hex());
+    let (mut nonce_out, mut rand_out) = ([0; 32], [0; 32]);
```
elichai, line 36: Why are you setting these parameters and then override them?

enigma-principal/app/src/esgx/epoch_keeper_u.rs
```diff
@@ -42,36 +53,51 @@ pub fn set_worker_params(eid: sgx_enclave_id_t, worker_params: &InputWorkerParam
     if retval != EnclaveReturn::Success || status != sgx_status_t::SGX_SUCCESS {
         return Err(EnclaveFailError { err: retval, status }.into());
     }
-    let seed = U256::from_big_endian(&rand_out);
-    let sig = Bytes(sig_out.to_vec());
-    let nonce = U256::from_big_endian(&nonce_out);
-    Ok(EpochState::new(seed, sig, nonce))
+    let epoch_state_out = match epoch_state {
```
elichai, line 56: Wait, so if the `epoch_state` already contains something than you just use it?
this means that use discarded the ecall data.

If so, why is that ecall made?

enigma-principal/app/src/esgx/epoch_keeper_u.rs
fredfortier: Which parameters are being overridden? I'll add more documentation in the code about the nonce and seed. In a nutshell, the purpose of `nonce_in` and `seed_in` is to restore the state of an active epoch. This is necessary to allow the program to resume after an outage. If the optional `EpochState` is given, the enclave verifies the epoch state derived from the inputs match the sealed epoch marker (a hash of the epoch state served by the enclave). If the optional state is not given (most of the time), it transitions to a new epoch and returns the seed and nonce. 

If you want, I can split this in two: `set_worker_params` and `verify_worker_params` to better separate concerns. I think that it would clarify the intent. The only reason why I didn't was to limit the number of ecalls.

enigma-principal/app/src/esgx/epoch_keeper_u.rs
fredfortier: Same as above, the ecall outputs (seed and nonce) are necessary to transition the `epoch_state`. When the optional `EpochState` is given, these attributes are expected to be equal so I agree that returning them redundant. That said, the ecall is still necessary to verify the inputs. I'm happy to split the ecall in two as suggested above.

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Doesn't `expect` panic just like `unwrap`? I want it do continue of the directory is already removed (possibly by another test running concurrently).

enigma-principal/enclave/src/epoch_keeper_t/epoch_t.rs
moriaab: @elichai @fredfortier 
Definitely KM node should use a general core implementation of hashing for multiple values, and not use a special function for specific values. The problem is that today in core there is no correct implementation of hashing of multiple values that may include nested arrays. Also the @fredfortier's implementation of `raw_encode` is not good as it may lead to the same output for different inputs. I and @elichai found an encoding for nested arrays, for which I did a formal proof that for any two different inputs, the outputs will be always different: https://github.com/enigmampc/protocol-discovery/blob/master/docs/hash_mul_nested.pdf. 
So it seems to me that:
 1. The correct implementation should be added to core
 2. KM node should use this implementation
 3. The implementation of hashing in the contract for signature verification should be changed accordingly.

Meanwhile before 1 is done, I am OK **to leave** `raw_encode` as is, because signature verification works now, and we can progress with integration tests and other stuff. Once 1 is done, we will do 2 and 3 ASAP.

enigma-principal/app/src/epoch_u/epoch_provider.rs
elichai: Oh. ok

elichai (dismissed):

enigma-principal/enclave/src/epoch_keeper_t/epoch_t.rs
fredfortier: > Also the @fredfortier's implementation of raw_encode is not good as it may lead to the same output for different inputs.

@moriaab Do you mind providing an example? This isn't my implementation, I implemented exactly what I was told the algorithm was in order to match the contract.

lacabra (approved):
Approved on the basis of @elichai's previous approval. Fred and I noticed the addition of 3 very old commits, and we were trying to remove them by changing the base branch and change it back to develop. That did not change those commits, but reset the approval.

Issue #124, opened 2019-04-18
Change Delta format in output of Deploy and Compute from String to Vec<u8>


lacabra:
Similar to #123, the output of a successful deploy currently looks like:
```
Ok(DeploySecretContract { result: DeployResult { 
pre_code_hash: "27245d9477ddbf32af803dd86c8bd94f5beed0a535e5efe989fbe63a2598a465",
used_gas: 6, output: "0061736d0100000...0000000000000",
delta: IpcDelta { contract_address: None, key: 0, data:
Some("90dd7c5499e164aa795b8dfeba398851f186a92223fd50e93beb0d300e4e36079c1c5dcbecb8238aa4af2b96c833fc0ce249ec776bda767b2bca0b35ed66e011f6") },
ethereum_address: "0000000000000000000000000000000000000000",
ethereum_payload: "", 
signature: "6de987fc2bdba11bc4367b317be5d837fc036ce927196c01332d679c6992c08c6e113366abc7979bb74005a26206e4039b7ee61cca3a576a1b77d378221e7b291c" } })
```
and the output of a successful compute currently looks like:
```
Ok(ComputeTask { result: ComputeResult { 
used_gas: 28077, 
output: "ef83228a5764098dc60bf6291d6a8c5637784b32ae31caa1b88e35bf99cdd256172ee319fe0f376c6ccc56d3ef2c266288527d66c495cb78f05ee2d7",
delta: IpcDelta { contract_address: None, key: 1, data:
Some("d3f292432e57ae3e3d14fc0f37bb84218ae5b543ddd9a5cc6728596054af2379d755ecc4ae4f5a61bd119944692a37a7837d55c4ed984f1fd5ec57126e2ca7caf3c9c7567af7033d2a163d7d2042fa9b174dd769201cf54750a76ccb6eda") },
ethereum_address: "0000000000000000000000000000000000000000",
ethereum_payload: "", signature: "71782688e7d1b923cb0f2c8345acd6b615849574ef99903cdaaf473928cf07955fa8e8d1fc51c71464214874ab0e8aea8304f9e26927e615ef2603d9adbd375c1c" } })
```

Note how the data in the deltas are hex strings, whereas they should be bytearrays. Again the P2P is all coded to process bytearrays in those fields, and receiving hex strings breaks the integrationTests.

[...]

PR #126, opened 2019-04-18
Changing of delta data type from String to Vec<u8>

moriaab:
https://github.com/enigmampc/enigma-core/issues/124


elichai (approved):
LGTM

PR #130, opened 2019-04-28
Fix the missing `rocksdb.so` in the docker

elichai:
I added `ldconfig` to the dockerfile so that the image will update it's links table for the dynamic linkage of rocksdb

### DIFF
--- a/dockerfile/Dockerfile
+++ b/dockerfile/Dockerfile
@@ -23,5 +23,7 @@ RUN git clone --depth 1  -b v1.0.4 https://github.com/baidu/rust-sgx-sdk.git  sg
 RUN git clone --depth 1 --branch v5.17.2 https://github.com/facebook/rocksdb.git rocksdb && \
     cd rocksdb && make install-shared && rm -rf /root/rocksdb
 
+RUN ldconfig
+
 RUN echo '/opt/intel/libsgx-enclave-common/aesm/aesm_service &' >> /root/.bashrc
 RUN echo 'export ROCKSDB_LIB_DIR=/usr/local/lib' >> /root/.bashrc


AvishaiW (approved):
IWOMM

elichai:
I'll rebase this ontop of https://github.com/enigmampc/enigma-core/pull/132 when fred looks at it, then the tests will pass.

lacabra:
FYI, there is a related issue to this, which I'm not sure this will fix or not, so I bring it up here for your attention:
- When you mount a core volume on the docker network and compile the code, the librocksdb file gets put in `/usr/local/lib`, but the files that are put in there are not persisted when you relaunch the docker network, so the next time you boot up the network with a mounted `core` volume you get an error.
- I temporarily fix it by manually copying those files inside a folder of the volume that I mount so that they are persisted, and amend the LD_LIBRARY_PATH pointing to that folder so that core takes them into account, but maybe the Makefile could put them into `enigma-core/lib` and we amend LD_LIBRARY_PATH to look for them there?

Or maybe you have other ideas on how to solve it.

elichai:
What do you mean they're not persistent? everything in the dockerfile should stay the same place unless overwritten

lacabra:
Yes, everything that is put in there at build time, but not what's changed at runtime

elichai:
Yeah, so it writes to `/usr/local/lib` in build time.

lacabra:
Yes, so if you mount the `core` repo onto `/root/enigma-core` that's the only thing that will persist. This only affects when you are doing development, and are mounting a volume, doing changes to that volume, and recompiling. Arguably, though, if you are only making changes to your code, the built librocksdb files shouldn't change, then you'll be fine. But if they change, and the new versions are put (at docker runtime) in `/usr/local/lib` those new versions will be lost when you bring down the network and up again.

elichai:
but how would they change? why would you touch `/usr/local/lib` at runtime?

lacabra:
> but how would they change? why would you touch /usr/local/lib at runtime?

As I mentioned above, this is only an issue for developers who are mounting a copy of the `core` repo. When you recompile core after making changes locally a new version of librocksbd may be generated, and put in `/usr/local/lib` by default

elichai:
> When you recompile core after making changes locally a new version of librocksbd may be generated, and put in `/usr/local/lib` by default

Did you observe a new `librocksdb` generated and put there? because that shouldn't happen. 
`rust-rocksdb` doesn't install librocksdb in the system.
it either dynamically link (this is the purpose of the `ROCKSDB_LIB_DIR` env), or compiles it locally and statically link it.
but it shouldn't "install" the lib anywhere outside of the project itself.

lacabra:
Thanks for the explanation. I can't tell for sure now. 

I suggest we proceed with this PR, which addresses an issue in its own right. I will then run some tests on the network and be in a position to answer your last set of questions above, and report back if there is an issue.

PR #132, opened 2019-04-28
fixes a bug that the KM node errors if `.enigma` doesn't exist

elichai:
@fredfortier this should fix your "warmup" problem.


fredfortier (approved):
This seems redundant because the creation of this subfolder is orchestrated in `esgx/general`. The problem is specific to unit tests. That said, I do not have any objection to using this approach.

elichai:
> This seems redundant because the creation of this subfolder is orchestrated in `esgx/general`. The problem is specific to unit tests. That said, I do not have any objection to using this approach.

I think that every function that creates a file should check if the directory already exists or not. otherwise it's untestable and can't be reused.

PR #135, opened 2019-04-29
Charge gas for writing into state, deployment and execution

moriaab:
I added gas charge for writing into state, deployment and execution.
1. The charging shceme  for writing into state is very simple, it is documented in comment for `Runtime::calculate_gas_for_writing`. @elichai please see if you agree with the scheme.
2. The gas for deployment is the gas for the wasm constructor execution + a component proportional to the size of the *deployed* code, which is known only *after* the constructor is executed.
3. The gas for execution is the gas for the wasm function execution + a constant value.
@elichai the code for the deployment and execution charge is divided between the `enigma-runtime-t/src/lib.rs` and enigma-core/enclave/src/wasm_g/execution.rs`, but it will be improved after the planned refactoring of runtime.


enigma-runtime-t/src/lib.rs
```diff
@@ -185,6 +205,36 @@ impl Runtime {
...
+   /// is subtracted from the initial value until 0.
+    fn calculate_gas_for_writing(&mut self, new_value_len: u32, key: &str) -> Result<u64> {
+        let mut result = 0;
+        let val = &self.post_execution_state.json[key];
+        let mut old_value_len = 0;
+        // forcing the length of Null value to be 0, since it is not 0.
+        if !val.is_null() {
+            let old_value_vec =
+                serde_json::to_vec(val).expect("Failed converting Value to vec in Runtime while reading state");
+            old_value_len = old_value_vec.len() as u32;
```
elichai: should we be worried about overflowing?
that someone could write a 2^32 bytes vector and will pay less?
or maybe it doesn't matter because he can't allocate that much anyway (even 2^32 bytes is 4GB)

p.s. we should have a test to see what happens in stack/heap overflow in the wasmi

enigma-runtime-t/src/lib.rs
```diff
@@ -185,6 +205,36 @@ impl Runtime {
...
+        }
+        if new_value_len >= old_value_len {
+            result = self.gas_costs.write_value + (new_value_len - old_value_len)*self.gas_costs.write_additional_byte;
+        }
+        else {
+            let decrease_cost = (old_value_len - new_value_len)*self.gas_costs.write_additional_byte;
+            if decrease_cost < self.gas_costs.write_value {
+                result = self.gas_costs.write_value - decrease_cost;
```
elichai: what if the `decrease_cost` is higher than write_value? should this remove from the total amount of gas? or just this gas?

in general idk about this separation of gases, but maybe this is something that should be handle when we redo the runtime.

elichai (approved):
Otherwise it looks good

enigma-runtime-t/src/lib.rs
moriaab: Wow, this is extremely important point, thanks! I think that we should check for overflow, even if it is not unlikely to happen. I created the task for checking the stack/heap overflow in wasmi.

enigma-runtime-t/src/lib.rs
moriaab: It's a good idea to decrease from the global gas. I changed the code accordingly.

moriaab:
@elichai please review the new commits

enigma-runtime-t/src/lib.rs
```diff
@@ -209,34 +210,62 @@ impl Runtime {
...
+        }
+    }
+
+    fn treat_gas_underflow(&mut self, val: &Option<u64>) -> (bool, u64){
+        let mut result = 0;
+        if val.is_none() {
+            self.gas_return = self.gas_limit;
+            (true, result)
+        } else {
+            (false, val.unwrap())
```
elichai: I think
```
if let Some(val) = val {
	(false, val)
}
```

Is better than unwrapping

enigma-runtime-t/src/lib.rs
```diff
@@ -209,34 +210,62 @@ impl Runtime {
...
-   /// If the new value is smaller than the old one, then the charge for each saved byte
-   /// is subtracted from the initial value until 0.
-    fn calculate_gas_for_writing(&mut self, new_value_len: u32, key: &str) -> Result<u64> {
-        let mut result = 0;
-        let val = &self.post_execution_state.json[key];
+   /// If the new value is larger than the old one, then gas is charged for the new bytes.
+   /// If the new value is smaller than the old one, then the gas is returned for the removed bytes.
+    fn calculate_gas_for_writing(&mut self, new_value_len: u64, key: &str) -> Result<u64> {
+        let mut result = Some(0);
+        let val = self.post_execution_state.json[key].clone();
```
elichai, line 243: Does it need to be cloned? I can't seem to see where it's used not by reference.

moriaab: self is used by reference

enigma-runtime-t/src/lib.rs
moriaab: I agree, I am changing this, thanks.

enigma-runtime-t/src/lib.rs
```diff
@@ -185,6 +210,64 @@ impl Runtime {
         Ok(())
     }
 
+    fn treat_gas_overflow(&mut self, val: &Option<u64>) -> Result<()>{
+        if val.is_none() {
+            self.gas_counter = self.gas_limit;
+            Err(WasmError::GasLimit)
+        }
+        else {
+            Ok(())
```
elichai, line 219: Maybe match over the `val` here and return the contained value in the `Ok()` then you won't need to later unwrap the val.

enigma-runtime-t/src/lib.rs
```diff
@@ -185,6 +210,64 @@ impl Runtime {
...
+        // If the new value is larger than the old one, the gas should be charged
+        if new_value_len >= old_value_len {
+            let checked_val = (new_value_len - old_value_len).checked_mul(self.gas_costs.write_additional_byte);
+            self.treat_gas_overflow(&checked_val)?;
+            result = self.gas_costs.write_value.checked_add(checked_val.unwrap());
+            self.treat_gas_overflow(&result)?;
+        } // If the new value is smaller than the old one, the gas should be returned
+        else {
+            let decrease_cost = (old_value_len - new_value_len).checked_mul(self.gas_costs.write_additional_byte);
+            let (underflow, val) = self.treat_gas_underflow(&decrease_cost);
```
elichai: I'm not sure how I understand how is this underflowing. `old_value_len` is bigger than `new_value_len` so subtracting it will yield a positive number, and then you're multiplying.

PR #142, opened 2019-05-05
Fixing the CI errors

elichai:
- Updated to the new yml format, the old one failed mounting `isgx` on the new drone version.

- Locked `cbindgen` to `0.8.4` (https://github.com/eqrion/cbindgen/issues/335)
- Locked `openssl-sys` to `0.9.44` (need to update `rustc` https://github.com/sfackler/rust-openssl/pull/1103 )
- Locked `rust-rocksdb` to `0.12.1` (https://github.com/rust-rocksdb/rust-rocksdb/issues/304)


.drone.yml
```diff
@@ -1,55 +1,65 @@
-pipeline:
-  core:
-    group: tests
+kind: pipeline
+name: default
+steps:
```
moriaab, line 3: As far as I understand, "steps" will lead to each step to be executed sequentially. If so, why is the advantage over "group"?

elichai: Yes, so everything I said about the yml today is deprecated lol
This is a new format, and yes steps are usually sequencial but adding the `depemds_on=[clone] ` makes them parallel.

moriaab: depends_on=[clone] was my next question. This new .yml format is so non-intuitive :)))

elichai: Yeah, they made it worse.

.drone.yml
```diff
@@ -1,55 +1,65 @@
...
       - . /root/.cargo/env
       - cd enigma-crypto
       - RUSTFLAGS=-Awarnings cargo test
       - RUSTFLAGS=-Awarnings cargo build --no-default-features --features=sgx
 
-  tools_m:
-    group: tests
-    image: enigmampc/enigma-core:0.0.5
+  - name: tools_m
+    image: enigmampc/enigma-core:0.0.5-rocksdb-5.17.2
```
moriaab, line 54: Did you catch the wrong docker image version by accident?

moriaab (approved):
Well done! It was not trivial as drone changed the configuration syntax.
@AvishaiW I suggest you to go through the changes in drone.yml

PR #143, opened 2019-05-06
BUG: change error print from debug to display which will show only th…

AvishaiW:
…e error without the backtrace. fixxes https://github.com/enigmampc/discovery-integration-tests/issues/10


moriaab (approved):
Important catch @AvishaiW @elichai!

PR #147, opened 2019-05-06
Adjusted the granularity of enclave logs

fredfortier:


enigma-principal/app/src/main.rs
```diff
@@ -43,27 +46,16 @@ mod esgx;
...
-            r
-        }
-        Err(x) => {
-            println!("[-] Init Enclave Failed {}!", x.as_str());
-            return;
-        }
-    };
-
-    // run THE app
+    let enclave = esgx::general::init_enclave_wrapper().expect("[-] Init Enclave Failed");
```
moriaab, line 55: This `main` is very similar to [that](https://github.com/enigmampc/enigma-core/blob/develop/enigma-core/app/src/main.rs#L19). This is not the only place where the code is copied for KM node. I think it worth time to go over the code and extract all the duplicate code to the common services **ASAP** (in a separate PR). Otherwise, we will get all the problems of the duplicate code (http://www.informit.com/articles/article.aspx?p=457502&seqNum=5)

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -52,7 +52,7 @@ fn get_state_keys(keys_map: &mut HashMap<ContractAddress, StateKey>,
                 println!("State key not found in cache, fetching sealed document.");
```
moriaab, line 51: Shouldn't the print in line 52 move to debug as well?

moriaab: The same question about lines 98, 99.

moriaab (approved):
I appreciate the reasonable size of this PR!
The PR looks OK. I left some comments. I would like to have your opinion especially on the comment about the duplicate code.

enigma-principal/app/src/main.rs
fredfortier: The principal node was initially all copied code. Of course I'm aware that code duplication is bad, but I don't feel comfortable modifying common components in this particular project. Fortunately, Elichai does and he did refactor most of the duplicate code into `enigma-tools-m`. The main file still has duplicated instructions but it's largely limited to initializing shared services. I would love to remove those duplicate instructions but I suggest that we  first discuss verbally and design this together.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: I thought preferable to leave some key INFO messages to indicate when new keys are being created. That said, I'm happy to make everything DEBUG if you prefer.

moriaab: This was only a question. I have no strong opinion about what should be DEBUG in this code and what should not. I trust your decision.

PR #148, opened 2019-05-07
TST: add test for validity of ethereum callback payload

moriaab:
Added a general test function for creation of Solidity ABI payload using ethabi interface and the contract ABI definition file.
The created payload is compared against the payload, which is the result of a secret contract execution.


AvishaiW (approved):
LGTM

PR #149, opened 2019-05-07
Wrapped Errors into Types

fredfortier:
Wrapped all errors into error types for more granular error handling by the client


lacabra (commented):
Works as desired in the integration tests in conjunction with p2p's [Issue 188](https://github.com/enigmampc/enigma-p2p/pull/188).

lacabra:
I defer to @elichai for approval

moriaab:
@fredfortier

moriaab:
@fredfortier @lacabra 
This PR raises a question regarding error handling. I don't understand how the errors of KM node are treated and passed outside. I see in `main` that if error is returned, the unwrap is done, which means that there will be panic. I didn't see any mapping of the errors to results of the execution. Does the error trace just go outside? To be more precise, how is the response [here](https://github.com/enigmampc/enigma-p2p/pull/188/files) created?

enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -34,7 +35,10 @@ impl TryInto<[u8; 65]> for StringWrapper {
     fn try_into(self) -> Result<[u8; 65], Self::Error> {
         let bytes = self.0.from_hex()?;
         if bytes.len() != 65 {
-            bail!("Cannot create a 65 bytes array from mismatching mismatching size slice.")
+            return Err(RequestValueErr {
+                request: METHOD_GET_STATE_KEYS.to_string(),
+                message: "Cannot create a 65 bytes array from mismatching mismatching size slice.".to_string(),
```
moriaab: Minor: _mismatching_ appears twice.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -98,14 +100,16 @@ impl EpochProvider {
         Ok(())
     }
 
-    #[logfn(DEBUG)]
+    //    #[logfn(DEBUG)]
```
moriaab: Is this comment left here on purpose?

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -98,14 +100,16 @@ impl EpochProvider {
...
     fn parse_worker_parameterized(&self, receipt: &TransactionReceipt) -> Result<Log, Error> {
         let log = receipt.logs[0].clone();
         let raw_log = RawLog { topics: log.topics, data: log.data.0 };
         let event = WorkersParameterizedEvent::new();
         let result = match event.0.parse_log(raw_log) {
             Ok(result) => result,
-            Err(_) => bail!("Unable to parse WorkersParameterized"),
+            Err(err) => return Err(Web3Error {
+                message: format!("Unable to parse WorkerParameterizedEvent: {:?}", err),
```
moriaab: As far as I understand the name of the event is `WorkerParameterized`, so I suggest to change `WorkerParameterizedEvent` to `WorkerParameterized event`.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
```diff
@@ -152,9 +153,10 @@ pub(crate) fn ecall_get_enc_state_keys_internal(
     for sc_addr in sc_addrs.clone() {
         let worker_addr = ecall_get_epoch_worker_internal(sc_addr)?;
         if worker_addr != recovered_addr {
-            return Err(SystemError(KeyProvisionError {
-                err: format!("Selected worker for contract: {:?} is not the message signer {} != {}",
-                             sc_addr, worker_addr.to_hex(), recovered_addr.to_hex())
+            println!("Recovered message signer address with:\nHash image: {}\nHash: {}", image.to_hex(), image.keccak256().to_hex());
```
moriaab: How this print is related to this condition `worker_addr != recovered_addr`?

moriaab:
General comment.
You have string constants in code like [that one](https://github.com/enigmampc/enigma-core/blob/74e6f6fbbf2c638c29ad9cf652814b950960a93e/enigma-principal/app/src/boot_network/keys_provider_http.rs#L20), but they are not always used as in [this case](https://github.com/enigmampc/enigma-core/blob/74e6f6fbbf2c638c29ad9cf652814b950960a93e/enigma-principal/app/src/cli/app.rs#L80).
Or there are names that are used multiple times through the code, but no constant is defined, for example `WorkerParameterized`:
https://github.com/enigmampc/enigma-core/blob/74e6f6fbbf2c638c29ad9cf652814b950960a93e/enigma-principal/app/src/boot_network/principal_manager.rs#L360
https://github.com/enigmampc/enigma-core/blob/74e6f6fbbf2c638c29ad9cf652814b950960a93e/enigma-principal/app/src/epoch_u/epoch_provider.rs#L110
Maybe it is worth walking through the code and changing this.

moriaab (approved):

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: No, I commented it out to see a stack trace. Will fix.

enigma-principal/enclave/src/keys_keeper_t/mod.rs
fredfortier: Victor asked for this to help troubleshoot this particular error. The address is recovered from the hash so I suppose that he wants to compare hash images to make sure that both systems are using the same data.

enigma-principal/app/src/boot_network/keys_provider_http.rs
fredfortier: Good catch. Will fix.

fredfortier:
@moriaab I try to always use constants for strings variables that are used in the code. In this case, the `WorkerParameterized` label is only used in the logs. I don't typically create new constants for log messages but I'm happy to change that if it is helpful. The `WorkerParameterized` event definition itself is defined as a type and reused throughout.

fredfortier:
@moriaab As far as know, all exceptions are checked. This allows the JSON-RPC server to create a response containing the error chain (although it does not seem like we have chained exceptions). It responds to each failed request with a generic error code + the exception thrown. For example, a response might look like this: `Unable to get keys: EpochStateTransitionErr { current_state: "UNCONFIRMED" }` or that `SystemError(WorkerAuthError { err: "Selected worker for contract: ... is not the message signer 2c3193fcbbe2980fbfc7a190505c1aa96a60bfbb != a098396d0cf4a0793cfba96dee60e4582752a281" })`. These new error types specify the cause with enough granularity for remediation. For example, an `EpochStateTransitionErr` indicates that a new epoch is not yet confirmed, so the client should try again later. A `WorkerAuthError` indicates the request signer is not authorized to get the requested keys, so retrying will not help.

enigma-principal/app/src/boot_network/keys_provider_http.rs
fredfortier: Fixed 5d5f96ec009007a8b5788cbc05cf4cbc78817c9e

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Fixed 4b73d61cb574b89eb255b1e9fde93d5fb34d383a

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Fixed 197978cebf824af8354d98bdbcc0fb3fe4f328e9

enigma-principal/enclave/src/keys_keeper_t/mod.rs
moriaab: In this case I think it is better to have a comment here explaining that this is for debug purposes and will be removed later.

moriaab:
Regarding string constants, it was a comment for your consideration, not a request for change. I think that constants are helpful even for log messages, but it's up to your decision.

PR #154, opened 2019-05-12
Refs #153

### LINKED ISSUE #153: KM node panics if signature is bad length
**Describe the bug**
The KM node panics if it receives a signature with valid hex bad bad length.
The error is at this line: https://github.com/enigmampc/enigma-core/blob/c18f65c11ff86d297776ca13b4a6cbe7507bd7ee/enigma-principal/app/src/boot_network/keys_provider_http.rs#L71
We pass the signature to `copy_from_slice` without checking length first which then asserts their length.

The solution is to implement something like this: https://github.com/enigmampc/enigma-core/blob/develop/enigma-types/src/hash.rs#L15

**To Reproduce**
Send this request to the KM node:
```
{ 
	data: "83a46461746181a752657175657374c0a269649c2d71cc8750cce9cca73c7cccd8ccbe73ccc6a67075626b6579dc0040224540cce8ccb6063118317e04cc9accb6741c17665ecca1ccd90106cc9acc8d0770cc9506cced323acce52ccc9fcca6cce465cce7cc e8cc9a33chdb39ccaf71ccee6557ccbccccc100eccfc4839cca2cce6cc9f4ecc88ccbb572725",
	sig: "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" 
}
```

**Expected behavior**
Return Invalid length error without crashing.

**Backtrace**
```
thread 'jsonrpc-eventloop-1' panicked at 'assertion failed: `(left == right)`
  left: `65`,    a      [                                  
 right: `32`: destination and source slices have different lengths', libcore/slice/mod.rs:1785:9
note:ERun with `RUST_BACKTRACE=1` for a backtrace.         

```

elichai:
This fixes #153 
And removes the unnecessary Backtrace from the JSON-RPC by using Display instead of Debug.


fredfortier (approved):
Looks good.

PR #155, opened 2019-05-12
Moved duplicate tools code to the mutual tools crate

elichai:
`LockExpectMutex` trait was duplicated for both `enigma-tools-u` and `enigma-tools-t`.
Moved them to `enigma-tools-m`.

I had a manual implementation of `ToHex` and `FromHex` as a legacy, I removed it and used `rustc-hex` from crates.io


enigma-tools-m/src/common/utils.rs
```diff
@@ -2,6 +2,21 @@ use crate::localstd::string::String;
...
+#[cfg(feature = "std")]
+use crate::localstd::sync::{Mutex, MutexGuard};
+
+
+pub trait LockExpectMutex<T> {
+    fn lock_expect(&self, name: &str) -> MutexGuard<T>;
+}
+
+impl<T> LockExpectMutex<T> for Mutex<T> {
+    fn lock_expect(&self, name: &str) -> MutexGuard<T> { self.lock().unwrap_or_else(|_| panic!("{} mutex is poison", name)) }
```
moriaab, line 17: I know that this comes from the original code and is not related directly to this refactoring, but it is important enough. Do we want `panic` here and not SystemError?

elichai: that's a good question.
generally as long as we don't have threads this will never panic.

this panics only if the mutex is poisoned, and the only way it can get poisoned is by a thread holding the lock panics without releasing it.

btw, you usually do want to panic here to propagate the panic between the threads.

https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning

moriaab (approved):

enigma-tools-m/src/common/utils.rs
moriaab: Ok, I see. It proves again how our code even conceptually is not ready for the multi-threading.

Issue #140, opened 2019-05-02
The principal node returns Error when it's tried too many times


elichai:
**Describe the bug**
When I send 100 requests at the same time to the principal node I get an error around 1 in 10 times.

**To Reproduce**
I'll post a code that spams the principal node with requests.

**Expected behavior**
Shouldn't return an error. either should *wait* for the mutex to be unlocked or change the design.

**Backtrace**
```
Err(Error(JsonRpcError(Error { code: InternalError, message: "Unable to get keys: ErrorMessage { msg: \"Unable to lock Epoch Marker Mutex.\" }", data: None }), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } }))
```
@fredfortier

fredfortier:
Thanks! It would be helpful if you could post your client.

AvishaiW:
@fredfortier what is the status on this issue? Did you manage to get the script from elichai?

fredfortier:
I think that this a duplicate of another bug that was fixed by changing the mutex `try_lock` to `lock`. AFAIK - this is fixed.

PR #156, opened 2019-05-14
Select Epoch By BlockNumber

fredfortier:
This PR modifies the epoch state management to store N epoch and make them available for worker selection verification. 

Here are some key requirements (also documented in EP-648):

1. The number of epoch stored should be configurable
2. The epoch state will be stored in the untrusted module and verified in the enclave. This extends the current sealed `EpochState` hash model and has been mostly implemented in issue EP-646 .
3. The JSON-RPC endpoint will contain a new optional `blockNumber` parameter corresponding to the block number of a Task Record. The Key Mgmt node will select the corresponding epoch if available. If not specified, the last confirmed epoch will be used.


elichai:
I'm not sure if this should be configurable from the untrusted side. because this is a consensus critical part.
I think this "configuration" should be hard coded in the enclave itself.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -56,146 +54,255 @@ impl EpochProvider {
...
-        let epoch_state = match File::open(Self::get_state_file_path()?) {
+    fn read_from_file(cap: usize) -> Result<Vec<EpochState>, Error> {
+        let epoch_state = match File::open(Self::get_file_path()?) {
             Ok(mut f) => {
                 let mut buf = Vec::new();
                 f.read_to_end(&mut buf)?;
                 let mut des = Deserializer::new(&buf[..]);
-                let epoch_state: Option<EpochState> = match Deserialize::deserialize(&mut des) {
-                    Ok(value) => Some(value),
+                let mut data: Vec<EpochState> = match Deserialize::deserialize(&mut des) {
```
elichai: btw, here you can use `unwrap_or_default()` which will make it way more readable.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -56,146 +54,255 @@ impl EpochProvider {
...
+    pub fn last(&self, exclude_unconfirmed: bool) -> Result<EpochState, Error> {
+        let guard = self.lock_guard_or_wait()?;
+        let mut epoch_state_val: Option<EpochState> = None;
+        for epoch_state in guard.iter().rev() {
+            if (exclude_unconfirmed && epoch_state.confirmed_state.is_some()) || !exclude_unconfirmed {
+                epoch_state_val = Some(epoch_state.clone());
+                break;
+            }
+        }
+        mem::drop(guard);
```
elichai: you can do just `drop(guard)` no need to call the `mem` module.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -56,146 +54,255 @@ impl EpochProvider {
...
+        match epoch_state_val {
```
elichai: you can do `epoch_state_val.ok_or(EpochStateUndefinedErr{})`
(you might need to add `into()` on the error or do `Ok(epoch_state_val.ok_or(EpochStateUndefinedErr{})?)` which will do it for you)

fredfortier:
@elichai I agree. There isn't much other choice anyway other than passing the capacity in the ecall which seems pointless. I skipped the cache capacity in the enclave for now but I'll implement it as a constant as suggested. I suppose that the untrusted component could read the capacity from the enclave to avoid duplicating this setting. But the benefits do not seem to warrant a new ecall so, unless you tell me otherwise, I'll leave the untrusted config as-is.

elichai:
I didn't understand what you're saying.
Should the untrusted part even be aware of the capacity?
If so(to check by itself before passing to the enclave or something like that) the capacity shouldn't be passed to the ecall and could be hard coded just like the enclave

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Fixed 24a61b09ab7ab77aa12f6722aa7f23cae3ad514d

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Fixed f6e788c409f21202784d047c5dc3825795581453

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Fixed fa9ed2c39c3c56a8c1fc52c2269aab08e3f75487

fredfortier:
Yes, I agree that the capacity should be hardcoded in the enclave. The untrusted part should be aware of capacity because it is keeping epoch data in cache (included worker selection mapping). Is there an area in the code where this constant can be declared once and shared? Or do you recommend declaring it twice?

elichai:
maybe put `pub const EPOCH_CAP: usize = X;` in `enigma-types`

fredfortier:
Done

enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -46,10 +48,28 @@ impl TryInto<[u8; 65]> for StringWrapper {
...
+        let bytes = self.0.from_hex()?;
+        if bytes.len() != 32 {
+            return Err(RequestValueErr {
+                request: METHOD_GET_STATE_KEYS.to_string(),
+                message: "Cannot create a 32 bytes array from mismatching size slice.".to_string(),
+            }.into());
+        }
+        let mut slice: [u8; 32] = [0; 32];
+        slice.copy_from_slice(&bytes);
+        Ok(U256::from(slice.as_ref()))
```
elichai, line 64: there's no need to copy it into an array if you then AsRef it into a slice anyway

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -56,146 +54,245 @@ impl EpochProvider {
...
-        println!("Parsed the {} event: {:?}", WORKER_PARAMETERIZED_EVENT, result);
-        Ok(result)
-    }
-
-    /// Returns the `EpochState` stored in memory
-    pub fn get_state(&self) -> Result<EpochState, Error> {
-        let guard = match self.epoch_state.try_lock() {
+    /// Lock the `EpochState` list `Mutex`, or wait and retry
+    pub fn lock_guard_or_wait(&self) -> Result<MutexGuard<Vec<EpochState>>, Error> {
+        // TODO: retry if locked
```
elichai, line 95: why not just `lock()` instead of `try_lock()`? then you wait for it to unlock.
otherwise this will result the same as here: https://github.com/enigmampc/enigma-core/issues/140

elichai: and you don't need a special function for it. locking a mutex automatically wait for it to be available. https://doc.rust-lang.org/std/sync/struct.Mutex.html https://doc.rust-lang.org/std/sync/type.TryLockResult.html

elichai (approved):

enigma-principal/app/src/epoch_u/epoch_provider.rs
fredfortier: Got it. I misunderstood the purpose to `try_lock`. I assumed that it was the same as `lock` with handled exceptions. I was surprised that Mutexes in Rust didn't block and wait. I understand the distinction now after reading the doc more carefully. I'll submit this change in a separate PR.

PR #157, opened 2019-05-14
Updating `rust-sgx-sdk` to 1.0.7

elichai:
Hi,
Until now we used `rust-sgx-sdk` 1.0.4, this PR updates it to 1.0.7.
This comes with a rust compiler/toolchain update to:
toolchain: `nightly-2019-04-26-x86_64-unknown-linux-gnu`
rustc: `rustc 1.36.0-nightly (3991285f5 2019-04-25)`

this means that for a while we won't need to lock more libraries, and we have some nice new features.

I also published a new docker to docker hub, version `0.0.7` that should do everything that is needed for here.

This work has been done in collaboration with @moriaab and @AvishaiW.


AvishaiW (approved):
fine

PR #158, opened 2019-05-14
Removed version locks and resolved warnings

elichai:
Now that we updated the rustc version (#157) We can remove the version locks.
I did that and removed a lot of warnings we had for a long time.

@moriaab / @AvishaiW every part is organized into a commit to help reviewing (I don't rewrite the same stuff between commits (as long as I didn't screw up the interactive rebasing lol))

### DIFF
--- a/enigma-core/app/src/db/dal.rs
+++ b/enigma-core/app/src/db/dal.rs
@@ -377,7 +377,7 @@ mod test {
     #[test]
     #[should_panic]
     fn test_fail_cf_exists_no_key_read() {
-        let (mut db, _dir) = create_test_db();
+        let (db, _dir) = create_test_db();
 
         let arr = [3u8; 32];
         let _cf = db.database.create_cf(&arr.to_hex(), &db.options).unwrap();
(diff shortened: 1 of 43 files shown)


enigma-core/app/src/db/dal.rs
```diff
@@ -377,7 +377,7 @@ mod test {
     #[test]
     #[should_panic]
     fn test_fail_cf_exists_no_key_read() {
-        let (mut db, _dir) = create_test_db();
+        let (db, _dir) = create_test_db();
```
AvishaiW, line 380: in tests that we don't modify the db, is it necessary to add it at all?

AvishaiW (approved):

elichai:
The problems in the tests were these: https://github.com/enigmampc/enigma-core/compare/358c7cdb00561193d64f9ab1e1815b267bda9660..5747449a31c33810f129b2bb6ff4eb152960ddff

elichai:
The problems in the tests were these: https://github.com/enigmampc/enigma-core/compare/358c7cdb00561193d64f9ab1e1815b267bda9660..6786720e2494c19353f047be210e5bad183ff449

(the eloop together with the underlying transport got dropped too soon and the `NODE_URL` wasn't set correctly)

PR #159, opened 2019-05-19
Remove manual serde Serialize/Deserialize implementation

elichai:
After Baidu's serde-derive will be updated to ^1.0.90 we can merge this and get rid of the manual serde implementations.

Thanks to: https://github.com/serde-rs/serde/pull/1499


AvishaiW (approved):
LGTM

PR #164, opened 2019-05-24
Documenting makefile

moriaab:
1. I documented makefile in a way that hopefully will be clear to the developers that are not familiar with makefiles.
2. I fixed `clean` command: enigma-core-app was not deleted by it.


enigma-core/Makefile
```diff
@@ -1,11 +1,21 @@
-######## My portable ############
+######## Portable Settings############
+
+# The path to baidu code. Uses environment variable HOME. In Enigma docker HOME=/root and sgx is cloned to /root.
```
AvishaiW:
```suggestion
# The path to baidu code. Uses environment variable HOME. In Enigma docker HOME=/root and sgx are cloned to /root.
```

enigma-core/Makefile
```diff
@@ -44,20 +56,33 @@ endif
 
 ######## CUSTOM Settings ########
 
+# Path with all libraries that are linked into enclave DLL
```
AvishaiW:
```suggestion
# Path of all libraries that are linked into enclave DLL
```

AvishaiW (commented):
thanks for adding the docs!

AvishaiW (approved):

PR #165, opened 2019-05-28
Auto generate ecalls FFI rust code using bindgen in enigma-core app

elichai:
This auto generates the FFI bindings in the app's `build.rs` file.

PR #166, opened 2019-05-28
Auto generate ecalls FFI rust code using bindgen in enigma-core enclave

elichai:
Because `build-dependencies` and regular dependencies are mixed together it's not possible to import `bindgen` into the enclave's `build.rs` (https://github.com/rust-lang/cargo/issues/2589)
The solution is to install the `bindgen` CLI in the docker and use it manually in the Makefile.


enigma-core/enclave/Makefile
```diff
@@ -13,4 +19,12 @@ ifeq ($(XARGO_SGX), 1)
...
\ No newline at end of file
+endif
+
+
+.PHONY: bindgen
+bindgen: Enclave_t.h
+	cargo build -p enigma-types # Meant to make sure `enigma-types.h` already exists and can be included.
+	echo $(BINDGEN_RAW_LINES)
+	bindgen Enclave_t.h $(BINDGEN_FLAGS) -- $(BINDGEN_CLANG_FLAGS) > $(BINDGEN_OUTPUT_FILE)
+	rustfmt  $(BINDGEN_OUTPUT_FILE)
```
moriaab, line 32: Why didn't you use rustfmt option of builder in  #165?

enigma-core/enclave/Makefile
```diff
@@ -13,4 +19,12 @@ ifeq ($(XARGO_SGX), 1)
...
 	cp ./target/$(Rust_target_dir)/libenigmacoreenclave.a ../lib/libenclave.a
-endif
\ No newline at end of file
+endif
+
+
+.PHONY: bindgen
+bindgen: Enclave_t.h
+	cargo build -p enigma-types # Meant to make sure `enigma-types.h` already exists and can be included.
+	echo $(BINDGEN_RAW_LINES)
```
moriaab: What is the reason for this echo?

moriaab (approved):
@elichai Please put the explanation of this PR as a comment in the code

enigma-core/enclave/Makefile
elichai: opssss

enigma-core/enclave/Makefile
elichai: I'll recheck but I'm pretty sure that it doesn't run on the added `raw_lines` which I add them all at the same line, so `rustfmt` will sort them later nicely

PR #167, opened 2019-05-28
Replacing the secp256k1 library

elichai:
This is more of a PoC for @AvishaiW or @moriaab to do in the future.
~~this should wait till after https://github.com/rust-bitcoin/rust-secp256k1/pull/115~~

https://enigmaco.atlassian.net/browse/EP-612

~~Replacing libsecp256k1 with rust-secp256k1 makes it not compile to WASM.
we should either investigate it further, Or feature gate between them
(another possibility is to outright not use it in wasm and maybe provide other crypto in a wasm specific program or provide these as external in the runtime)~~

### DIFF
--- a/enigma-core/app/Cargo.lock
+++ b/enigma-core/app/Cargo.lock
@@ -31,11 +31,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arrayref"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "arrayvec"
 version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


elichai:
FYI https://github.com/rust-bitcoin/rust-secp256k1/blob/master/CHANGELOG.md#0140---2019-07-08
@AvishaiW @moriaab 

P.S. maybe moving the Context to a global `lazy_static!` would be better. because the context is a one time slow creation that then makes everything really fast, and it's thread safe, so no need even for a Mutex. 
(so either lazy_static, or something like: https://github.com/elichai/ecc-secp256k1/blob/master/src/secp256k1.rs#L480)

moriaab:
Thank you so much @elichai

PR #168, opened 2019-05-30
DEV: Implemented new JSON-RPC error codes

fredfortier:
New JSON-RPC error codes:

- `-32001`: Worker not authorized to get the requested keys. The worker selection algorithm returned a negative. This means that the worker is either dishonest or somehow not in sync.
- `-32002`: Illegal epoch state. The state is transitioning and the `setWorkerParams` tx hasn't been confirmed yet.


moriaab (approved):
Nice solution with the jsonrpc error codes!
1) I would suggest to extract the lines 151-186 to a separate function that deals with errors for the reason that this code is not a part of the main logic, but it is long and distracting.
2) I think that this is very important to make constants for jsonrpc implementation-defined server errors:
    - meaningful names will explain the errors better. 
    - this way it will be easier to track the constants, since more errors can be added in the future.
I would also add a comment about the predefined errors range, because the numbers are not so intuitive :)

fredfortier:
Added the constants requested.

moriaab:
@fredfortier 
Thank you for the changes. Regarding the comment in the code, I think that the chosen idiom is good. Please, merge the PR.

PR #170, opened 2019-06-05
DEV: Using `lock` instead of `try_lock` 

fredfortier:


AvishaiW (approved):
Thanks for opening a new PR! LGTM

PR #171, opened 2019-06-05
Added an IPC Deployment message

AvishaiW:
a proposed fix on core for https://github.com/enigmampc/discovery-docker-network/issues/21

### DIFF
--- a/enigma-core/app/src/networking/ipc_listener.rs
+++ b/enigma-core/app/src/networking/ipc_listener.rs
@@ -236,10 +236,6 @@ pub(self) mod handling {
     #[logfn(INFO)]
     pub fn update_new_contract_on_deployment(db: &mut DB, address: String, bytecode: &str, delta: IpcDelta) -> ResponseResult {
         let mut tuples = Vec::with_capacity(DEPLOYMENT_VALS_LEN);
-        let delta_address = delta.contract_address.ok_or(P2PErr { cmd: "UpdateNewContractOnDeployment".to_string(), msg: "Address Missing".to_string() })?;
-        if address != delta_address {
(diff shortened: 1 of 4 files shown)


enigma-core/app/src/networking/ipc_listener.rs
```diff
@@ -230,6 +233,28 @@ pub(self) mod handling {
...
+        let bytecode_delta_key = DeltaKey::new(address_arr, Stype::ByteCode);
+        tuples.push((bytecode_delta_key, bytecode));
+
+        let data = delta.data.ok_or(P2PErr { cmd: "UpdateNewContractOnDeployment".to_string(), msg: "Delta Data Missing".to_string() })?;
+        let delta_key = DeltaKey::new(address_arr, Stype::Delta(delta.key));
+        tuples.push((delta_key, data));
+
+        let results = db.insert_tuples(&tuples);
+        let mut status: i8 = 0;
+        for result in results {
```
moriaab: It is better to replace -1 with FAILED. In addition, IMO use of `any` would be nicer here. Something like that:
if results.into_iter().any(|&x| x.is_err()){
      status = FAILED;
}

enigma-core/app/src/networking/messages.rs
```diff
@@ -62,6 +63,8 @@ pub enum IpcResults {
     #[serde(rename = "result")]
     UpdateDeltasResult { status: Status, errors: Vec<IpcStatusResult> },
     #[serde(rename = "result")]
+    UpdateDeploymentResult { status: Status },
```
moriaab: Why don't you just use plain Status as for `update_new_contract`?

moriaab (approved):
This is not related to this PR, but I bring it here since it relates to the code area of the PR. 
I think that it is better to have an enum for the status, which will include FAILED. We have one in `enygma-types`, maybe it can be reused.

PR #172, opened 2019-06-05
MAINT: updated simple_addition contract

lacabra:
Maintenance commit. As per the comment on the removed line, needs to be removed as nightly was already updated. I am using this for `enigmampc/discovery-cli` as I believe it's the simpler contract we have, and I would like to pull it from here directly, as otherwise (until I submitted this PR) I was keeping a local copy in another repo that is redundant.


AvishaiW:
unfortunately, baidu's docker which we rely on uses an older nightly version ([20-05-2019](https://github.com/baidu/rust-sgx-sdk#v108-release)), therefore if this change is pushed we won't be able to use this when running on the docker.

lacabra:
Thanks @AvishaiW. Closing it without merging then.

PR #175, opened 2019-06-13
Changing contract tests to get eng-wasm and eng-wasm-derive from crates.io

moriaab:
This change causes contract tests to fail if one of the following is not updated in crates.io:
eng-pwasm-abi
eng-wasm
eng-wasm-derive


AvishaiW (approved):
Great!

PR #179, opened 2019-06-18
Moving log of bytecode returning from deploy_contract function from info to debug level

moriaab:
The result returned from `deploy_contract` function contains bytecode, which may be huge. I moved it from *info* to *debug* level.


enigma-core/app/src/networking/ipc_listener.rs
```diff
@@ -353,9 +351,16 @@ pub(self) mod handling {
                 // Save the ExeCode into the DB.
                 let key = DeltaKey::new(contract_address, Stype::ByteCode);
                 db.create(&key, &v.output)?;
-                Ok(v.into_deploy_response(&bytecode))
+                let wasm_result = v.into_deploy_response(&bytecode);
```
AvishaiW: `wasm_result` is not a good name since it contains an IpcResponse

enigma-core/app/src/networking/ipc_listener.rs
```diff
@@ -353,9 +351,16 @@ pub(self) mod handling {
...
+                info!("deploy_contract() => Ok({})", wasm_result.display_without_bytecode());
+                debug!("deploy_contract() => {}", wasm_result.display_bytecode());
```
AvishaiW:
```suggestion
                debug!("deployed bytecode => {}", wasm_result.display_bytecode());
```

enigma-core/app/src/networking/messages.rs
```diff
@@ -44,6 +44,39 @@ pub enum IpcResponse {
...
+            _ => "".to_string(),
+        }
+    }
+
+    pub fn display_bytecode(&self) -> String {
+        match self {
+            IpcResponse::DeploySecretContract {result: e} => {
+                match e {
+                    IpcResults::DeployResult {  output,.. } =>
+                        format!("bytecode: {}", output),
```
AvishaiW:
```suggestion
                        format!("{}", output),
```

enigma-core/app/src/networking/ipc_listener.rs
AvishaiW: I don't like that we have two separate outputs that contain the deploy_contract signature, it might be preferable to structure it in this way. just a thought, feel free to decide whatever..

AvishaiW (approved):
With the available circumstances, this looks great!

enigma-core/app/src/networking/ipc_listener.rs
moriaab: Done

PR #190, opened 2019-07-10
MAINT: adj verbosity after #179

lacabra:
Adjusting verbosity of the `core` container after #179, which had been reduced a level because the Travis CI builds were failing because of too much output. This seems the right level of verbosity now, if the Travis CI builds fail, we will adjust it elsewhere (and yes it's Travis when running the integration tests for the `contract` and `p2p`, and not Drone where the tests for this repo run).


AvishaiW (approved):
Great!

PR #187, opened 2019-07-03
Update rust-toolchain to nightly-2019-05-22

elichai:
`rust-toolchain` file doesn't match baidu's toolchain https://github.com/baidu/rust-sgx-sdk/commit/e0fa56e7deca38fdba3797b345a0a46b696edd8b

This should also fix the annoying `no_mangle` warnings https://github.com/rust-lang/rust/issues/60050

PR #191, opened 2019-07-10
updated the rust toolchain according to baidu and #187

AvishaiW:
according to @elichai's PR: #187


moriaab (approved):

AvishaiW:
@lacabra it seems that the reason for the failure in the CI was due to a duplicate download of rustfmt both in baidu's and our image. and since there were breaking changes in the `nightly-2019-05-22` toolchain, baidu downloaded the specific version for this toolchain but on our dockerfile, we didn't specify the exact version, therefore, we ran over the needed version and downloaded the latest available leading to the breaking code while running in the docker.

AvishaiW:
@lacabra, as I mentioned already, the reason for the failure, was the fact that we had a duplicate download for rustfmt which ran over the proper version that Baidu's image downloaded. I removed that line and updated the image version.

PR #196, opened 2019-07-18
removed shared dependencies in the tests which fail due to parallelism

AvishaiW:
added lots of refactoring as well.


enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -222,7 +222,7 @@ mod test {
 
     #[test]
     pub fn test_jsonrpc_get_state_keys() {
-        setup_epoch_storage();
+        let path = setup_epoch_storage();
```
moriaab: There is no use of this variable `path`. Is it needed?

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -99,19 +100,20 @@ impl ReportManager {
     }
 }
 
-impl PrincipalManager {
+impl PrincipalConfig {
```
moriaab, line 103: Nice!

enigma-principal/app/src/boot_network/keys_provider_http.rs
AvishaiW: removed it.

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -77,20 +78,20 @@ impl ReportManager {
     pub fn get_registration_params(&self) -> Result<RegistrationParams, Error> {
         let signing_address = self.get_signing_address()?;
         let sim_mode = option_env!("SGX_MODE").unwrap_or_default();
-        println!("Using SGX_MODE: {:?}", sim_mode);
-        println!("Fetching quote with SPID: {:?}", self.config.spid);
+        info!("Using SGX_MODE: {:?}", sim_mode);
```
moriaab: I am not sure under what conditions should the information of `get_registration_params` be printed. @lacabra, what do you think about this change?

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -30,64 +30,60 @@ use std::mem::replace;
...
     pub cap: usize,
+    pub dir_path: PathBuf,
 }
 
-impl EpochStateManager {
-    pub fn new(cap: usize) -> Result<Self, Error> {
-        let epoch_state_val = Self::read_from_file(cap)?;
-        let epoch_state_list = Mutex::new(epoch_state_val.clone());
-        Ok(Self { epoch_state_list, cap })
+fn get_file_path(mut path: PathBuf) -> Result<PathBuf, Error> {
```
moriaab: Can the path be a reference here? The argument is cloned in every call to this function, maybe it can be avoided.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -30,64 +30,60 @@ use std::mem::replace;
...
-        if epoch_state.is_empty() {
-            return Ok(fs::remove_file(path).unwrap_or_else(|_| println!("No epoch state file to remove")));
-        }
-        let mut file = File::create(path)?;
-        let mut buf = Vec::new();
-        epoch_state.serialize(&mut Serializer::new(&mut buf))?;
-        file.write_all(&buf)?;
-        Ok(())
+impl EpochStateManager {
+    pub fn new(mut dir_path: PathBuf, cap: usize) -> Result<Self, Error> {
```
moriaab: Why EpochStateManager holds directory and not file. Why it is always needed to pass through `get_file_path`?

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -211,13 +215,13 @@ impl Sampler for PrincipalManager {
     /// * `reset_epoch` - If true, reset the epoch state
     /// * `gas_limit` - The gas limit for all Enigma contract transactions
     #[logfn(INFO)]
-    fn run<G: Into<U256>>(&self, reset_epoch: bool, gas_limit: G) -> Result<(), Error> {
+    fn run<G: Into<U256>>(&self, path: PathBuf, reset_epoch: bool, gas_limit: G) -> Result<(), Error> {
```
moriaab, line 219: Please add a new argument to a documentation of the function

moriaab (approved):
It is a good catch!
I think that the functions `get_file_path`, `read_from_file`, `write_to_file` belong to EpochStateManager and should be about its inside file. I think that for the test `test_write_epoch_state, one such mock manager should be created.
If you decide that this is wrong and the functions should be global, at least their names should be changed to something less generic.

enigma-principal/app/src/boot_network/principal_manager.rs
lacabra: I would argue for Verbosity L2 (-vv), even L1 since that info is very useful, but in an upcoming change to P2P, that info should be displayed upon request from P2P, so the discussion should be moved there. But for now L1 or L2, if that makes sense. Thx!

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -212,6 +212,7 @@ impl Sampler for PrincipalManager {
     ///
     /// # Arguments
     ///
+    /// * `path` - path to the .enigma directory.
```
moriaab: Does this parameter always have the same value? If so , then this parameter is redundant.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -79,11 +70,16 @@ fn write_to_file(path: PathBuf, epoch_state: Vec<EpochState>) -> Result<(), Erro
 }
 
 impl EpochStateManager {
-    pub fn new(mut dir_path: PathBuf, cap: usize) -> Result<Self, Error> {
-        dir_path.push(EPOCH_DIR);
-        let epoch_state_val = read_from_file(dir_path.clone(), cap)?;
+    pub fn new(mut state_path: PathBuf, cap: usize) -> Result<Self, Error> {
```
moriaab, line 38: `EpochStateManager` holds `state_path`. Why it is not set here? 
I think that `write_to_epoch_state` and `read_from_epoch_state` should be functions of `EpochStateManager`. The signs for that:
- the names of these functions
- the functions get `EpochStateManager` as parameter
- the functions get path as a parameter, this path should be an attribute of the `EpochStateManager`

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -378,28 +379,59 @@ pub mod test {
...
-//        let sig = Bytes::from(mock_sig.to_vec());
-//        let nonce = U256::from(0);
-//        let epoch_state = EpochState { seed, sig, nonce, confirmed_state };
-//        write_to_epoch_state(&path, vec![epoch_state.clone()]).unwrap();
-//
-//        // TODO: This could fail if another test deleted the epoch files exactly here, give unique name
-//        let saved_epoch_state = read_from_epoch_state(&path, 1).unwrap();
-//        assert_eq!(format!("{:?}", saved_epoch_state[0]), format!("{:?}", epoch_state));
-//    }
+    #[test]
```
moriaab, line 386: Excellent!

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -33,52 +33,40 @@ pub struct EpochStateManager {
...
     pub fn new(mut state_path: PathBuf, cap: usize) -> Result<Self, Error> {
         state_path.push(EPOCH_DIR);
         if !state_path.exists() {
             fs::create_dir_all(&state_path)?;
         }
         state_path.push(EPOCH_FILE);
-        let epoch_state_val = read_from_epoch_state(&state_path, cap)?;
-        let epoch_state_list = Mutex::new(epoch_state_val.clone());
-        Ok(Self { epoch_state_list, cap, state_path })
+        Self::create_from_path(state_path, cap)
```
moriaab, line 44: Now it looks good.

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -378,28 +379,59 @@ pub mod test {
     pub fn setup_epoch_storage_file() -> PathBuf {
```
moriaab: Maybe it is better to call it `setup_epoch_storage_dir`?

PR #198, opened 2019-07-30
Adding a Symmetric Crypto Service to ewasm

fredfortier:
- [x] Public interfaces
- [x] Resolver
- [x] Implementation in enclave


eng-wasm/src/crypto_wasm.rs
```diff
@@ -0,0 +1,17 @@
+use super::*;
+
+const SYMMETRIC_KEY_SIZE: usize = 32;
```
moriaab: It is better to use already existing constants in symmetric.rs

enigma-core/enclave/src/auto_ffi.rs
```diff
@@ -1,8 +1,8 @@
-// automatically generated by rust-bindgen
+cargo:warning=couldn't execute `llvm-config --prefix` (error: No such file or directory (os error 2))
```
moriaab: Please remove this from the PR. The file's contents didn't change, but there wos a warning during the generation.

enigma-crypto/src/symmetric.rs
```diff
@@ -15,10 +15,14 @@ use crate::localstd::vec::Vec;
 use crate::localstd::vec;
 use crate::rand;
 
-const IV_SIZE: usize = 96/8;
 static AES_MODE: &aead::Algorithm = &aead::AES_256_GCM;
-type IV = [u8; IV_SIZE];
 
+/// The symmertic key byte size
+pub const SYMMETRIC_KEY_SIZE: usize = 32;
```
moriaab: What is the purpose of this constant in this file if it is not used in the code?

enigma-runtime-t/src/lib.rs
```diff
@@ -381,11 +378,48 @@ impl Runtime {
...
+        let mut key: SymmetricKey = [0u8; SYMMETRIC_KEY_SIZE];
+        self.memory.get_into(key_ptr, &mut key)?;
+
+        let iv_ptr: u32 = args.nth_checked(3)?;
+        let mut iv: IV = [0u8; IV_SIZE];
+        self.memory.get_into(iv_ptr, &mut iv)?;
+
+        let ptr: u32 = args.nth_checked(4)?;
+        let enc_message = encrypt_with_nonce(&message, &key, Some(iv))
+            .map_err(|err| WasmError::EnclaveError(EnclaveError::SystemError(EnclaveSystemError::CryptoError { err })))?;
```
moriaab: Please implement the need `From` for errors. We did an effort to keep the runtime readable and free from conversions :) See https://github.com/enigmampc/enigma-core/blob/develop/enigma-tools-t/src/common/errors_t.rs for the reference.

enigma-runtime-t/src/lib.rs
```diff
@@ -381,11 +378,48 @@ impl Runtime {
...
+        let cipheriv_len: u32 = args.nth_checked(1)?;
+        let cipheriv = self.memory.get(cipheriv_ptr, cipheriv_len as usize)?;
+
+        let key_ptr: u32 = args.nth_checked(2)?;
+        let mut key: SymmetricKey = [0u8; SYMMETRIC_KEY_SIZE];
+        self.memory.get_into(key_ptr, &mut key)?;
+
+        let ptr: u32 = args.nth_checked(3)?;
+        let message = decrypt(&cipheriv, &key)
+            .map_err(|err| WasmError::EnclaveError(EnclaveError::SystemError(EnclaveSystemError::CryptoError { err })))?;
```
moriaab: Please implement the need From for errors. We did an effort to keep the runtime readable and free from conversions :) See https://github.com/enigmampc/enigma-core/blob/develop/enigma-tools-t/src/common/errors_t.rs for the reference.

eng-wasm/src/crypto_wasm.rs
moriaab, line 6: Can these constants and the constants in symmetric.rs have the same single origin and not be copied? Duplication may cause big problems in the future, if only one of the places will be changed.

moriaab (commented):
Great job!
Please add tests for the usage of this feature.

eng-wasm/src/crypto_wasm.rs
fredfortier: Yes, but where would you put them? `eng_wasm` does not seem to use any of the common modules.

enigma-crypto/src/symmetric.rs
fredfortier: It's used in `runtime_t`. I'm just trying to avoid re-defining constants that can be imported. Let me know if there's a better place to put shared constants.

fredfortier:
@moriaab Where should I put the unit tests? I can't see to find any for the `eng_wasm` module.

enigma-crypto/src/symmetric.rs
fredfortier: Added directly in `runtime_t` for now. Please let me know what is the right place for this constant. I don't see any shared module between `eng_wasm` and `runtime_t`.

enigma-runtime-t/src/lib.rs
fredfortier: Done.

enigma-runtime-t/src/lib.rs
fredfortier: Done.

enigma-core/enclave/src/auto_ffi.rs
fredfortier: Done

eng-wasm/src/lib.rs
```diff
@@ -123,5 +127,10 @@ mod tests {
     fn what() {
         print("TEST!");
     }
+
+    #[test]
+    fn test_encrypt() {
+        // TODO: Is this the right place to test APIs. If so, how should we initialize the enclave?
```
moriaab, line 133: It is better to put the tests [here](https://github.com/enigmampc/enigma-core/blob/develop/enigma-core/app/src/wasm_u/wasm.rs#L61)

1. Create the contract that uses encrypt/decrypt services.
2. Put the contract in https://github.com/enigmampc/enigma-core/tree/develop/examples/eng_wasm_contracts
3. **In Cargo.toml of the contract use the local paths for `eng-wasm` and `eng-wasm-derive`** instead of crates from crates.io.
4. Create a test in wasm.rs. All the helper functions are there as well as examples of working with `ethabi::Token`. Use the existing tests for reference. Here is the simplest form of the test https://github.com/enigmampc/enigma-core/blob/develop/enigma-core/app/src/wasm_u/wasm.rs#L363.

fredfortier: Done.

eng-wasm/src/crypto_wasm.rs
```diff
@@ -15,10 +16,42 @@ pub fn generate_key() -> SymmetricKey {
     key.0
 }
 
-pub fn encrypt(message: &[u8], key: &SymmetricKey, payload: &mut [u8]) {
-    unsafe { external::encrypt(message.as_ptr(), message.len() as u32, key.as_ptr(), payload.as_ptr()) };
+pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
+    // TODO: Is this really needed? Dynamically sized buffers don't seem to work.
+    // TODO: Is it possible to estimate the encrypted payload size based on the plaintext message?
+    let length = PAYLOAD_SIZE;
+    let mut payload = Vec::with_capacity(length);
```
fredfortier, line 23: @moriaab I tried everything that I could to use a dynamic buffer but it only seems to work with a static buffer allocated with zeros. Please confirm that dynamic buffers are not supported. If true, I'll see if I can optimize by estimating the size of the encrypted data instead of using a constant.

enigma-core/app/src/wasm_u/wasm.rs
```diff
@@ -374,6 +374,22 @@ mod tests {
...
+            &mut db,
+            "../../examples/eng_wasm_contracts/encryption",
+            generate_contract_address(),
+            "construct()",
+            &[],
+            "encrypt_decrypt(bytes)",
+            &[Token::Bytes(message.to_vec())]
+        );
+    }
+
```
moriaab, line 392: The meaning of test is not just invoke the functionality, but to compare the results with the expected ones.

fredfortier: The purpose of this test is to verify encryption/decryption, not to transform a value and verify the output. The contract will panic if the value isn't decrypted correctly and the test will fail. 

Please understand that the reason why I didn't return an output is because it is redundant. You should not assume that I don't know what a test is. I simply used `test_print_simple` as a template as per your suggestion. If you prefer that each function invoked by `compile_deploy_execute` must always return an output, just tell me and I'll happily make that change.

moriaab: Oh, you are right, sorry, my comment was wrong.

examples/eng_wasm_contracts/encryption/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn encrypt(plaintext_msg: &Vec<u8>) -> Vec<u8> {
+        let key = Self::get_pkey();
+        eprint!("Encrypting bytes ({:?})", plaintext_msg);
+        encrypt(plaintext_msg, &key)
+    }
+}
+
+impl ContractInterface for Contract {
+    // Constructor function that takes in VotingETH ethereum contract address
+    #[no_mangle]
```
moriaab, line 76: I realized that `no_mangle` are meaningless in secret contracts, so I removed it from all our example contracts. Please remove `no_mangle` to avoid the confusion about this contract.
Please remove the comment part about `no_mangle`: lines 47,48.

examples/eng_wasm_contracts/encryption/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+// #[no_mangle] modifier is required before each function to turn off Rust's name mangling, so that
+// it is easier to link to. Sets the symbol for this item to its identifier.
+pub struct Contract;
+
+// Private functions accessible only by the secret contract
+impl Contract {
+    fn get_pkey() -> SymmetricKey {
+        let key = read_state!(ENCRYPTION_KEY).unwrap();
+        let key_pair = KeyPair::from_slice(&key).unwrap();
+        let pub_key = key_pair.get_pubkey();
```
moriaab, line 56: Can you please clarify what is the purpose of `pub_key`?

examples/eng_wasm_contracts/encryption/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+extern crate rustc_hex;
+
+// eng_wasm
+use eng_wasm::*;
+use eng_wasm_derive::pub_interface;
+use eng_wasm::{String, Vec};
+use rustc_hex::ToHex;
+use enigma_crypto::asymmetric::KeyPair;
+
+// State key name "mixer_eth_addr" holding eth address of Mixer contract
```
moriaab, line 34: The lines 27, 34 and 35 are commented out, if there is no potential use, maybe it is better to remove them.

eng-wasm/src/crypto_wasm.rs
moriaab: @fredfortier I think that maybe getting `SYMMETRIC_KEY_SIZE` and `IV` should be also part of runtime services? What do you think?

eng-wasm/src/crypto_wasm.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub fn generate_key() -> SymmetricKey {
+    let key_int: U256 = Rand::gen();
+    let key = H256::from(key_int);
+    key.0
+}
+
+pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
+    // TODO: Is this really needed? Dynamically sized buffers don't seem to work.
+    // TODO: Is it possible to estimate the encrypted payload size based on the plaintext message?
```
moriaab, line 21: @fredfortier I think that you can estimate the encrypted payload size. In block cypher it is equal to the plaintext message with completion of the last part to a block size.

eng-wasm/src/crypto_wasm.rs
```diff
@@ -0,0 +1,57 @@
...
+pub type SymmetricKey = [u8; SYMMETRIC_KEY_SIZE];
+
+pub fn generate_key() -> SymmetricKey {
+    let key_int: U256 = Rand::gen();
+    let key = H256::from(key_int);
+    key.0
+}
+
+pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
+    // TODO: Is this really needed? Dynamically sized buffers don't seem to work.
```
moriaab, line 20: @fredfortier As far as I understand, it is impossible to use the dynamic buffers here. You should allocate a continuous region of WASM memory. This region is filled when the runtime function is called. The dynamic array cannot be managed from the runtime.

moriaab (approved):
I added more comments, will appreciate if you could address them before merging. 
Thank you.

examples/eng_wasm_contracts/encryption/src/lib.rs
fredfortier: It does not do anything with local encryption. It is just verifying that it is possible to obtain valid key material from the generated private key. I do use this method in the coinjoin contract to encrypt data externally without sharing the private key.

eng-wasm/src/crypto_wasm.rs
fredfortier: Are you ok if I keep the static size for now and create a new PR to implement this? I other things to finish before I can revisit this and I don't want to keep the PR open for too long.

moriaab: Sure.
To make your merge easy, I am waiting for your merge before I merge PR#204, which changes the structure of the runtime.

PR #205, opened 2019-08-08
DEV: removing no_mangle

moriaab:
`no_mangle` directive does not affect the functionality of the program. The secret contract functions are addressed by a procedural macro `pub_interface` by their names. Compilation names does not matter.


AvishaiW (approved):
Amazing!

PR #209, opened 2019-08-27
Ep 748 simplify pub interface

reuvenpo:
rewritten eng_wasm_derive::pub_interface

* Enabled usage of the \#[pub_interface] annotation on `impl Struct` items
  as well as on trait items. When using the new form, the user may use whatever name they want
  for the struct. When using the trait version the name of the struct/enum implementing
  the methods was hardcoded as `Contract`, but is now customizable by passing the 
  implementor name as an argument to the macro.
  (They can basically use any type they want now)
* Added the following diagnostics to the macro invocation:
  (when any of these is detected, the user will see a useful error message)
    * The attribute was placed on an item which is not a trait or inherent impl
    * The trait or impl methods use the `self` receiver
    * Trait methods provided a default implementation (i understand that we
      wanted the users to define the implementations on the `Contract` struct)
    * Trait method with any additional attribute
    * Impl method with a \#[no_mangle] attribute
    * Impl method with visibility which isn't private or `pub`
      (we don't want them defining `pub(crate)` or `pub(in something)` because
      that can be misleading in this context)
    * check that the constructor returns no value
    * check that when applied to an impl-item, the constructor is marked as `pub`
* Fixed issue around parsing return values from contract methods where their
  amount was incorrectly calculated
* When parsing of method inputs fails at runtime, the panic message now
  mentions the type and argument that caused the failure.
* If an unknown method is called, the panic message mentions the name of
  that method.
* Using the macro no longer requires writing `use eng_wasm::*;` in the using
  module
* restructured the module layout to make it easier to follow.
* added unittests
* documented a lot of things

The ERC20 example now generates the following code:
https://gist.github.com/rust-play/d9e375f7a587725be238ced968630d28
EDIT:
https://gist.github.com/rust-play/4a1da16589b738ee2f206d351dfd4c84


eng-wasm/derive/src/eth_contract.rs
```diff
@@ -0,0 +1,154 @@
+use std::fs::File;
```
reuvenpo, line 1: I copied the implementation of the `eth_contract` macro to this new submodule. i did not otherwise change the implementation, except a small change related to breaking changes in `syn` 1.0

eng-wasm/derive/src/eth_contract.rs
```diff
@@ -0,0 +1,154 @@
...
+    let it: Vec<proc_macro2::TokenStream> = generate_eth_functions(&contract).unwrap();
+
+    quote! {
+        struct #struct_name {
+            addr: Address,
+        }
+        impl EthContract{
+            fn new(addr_str: /*Address*/&str) -> Self {
+                use core::str::FromStr;
+        // Ethereum Addresses need to start with `0x` so we remove the first two characters
```
reuvenpo: I should fix the indentation here

eng-wasm/derive/.rustfmt.toml
```diff
@@ -0,0 +1 @@
+# override global formatting. i want clean and standard runstfmt.
```
reuvenpo: Th project directory contains a fustfmt configuration file with a lot of unusual settings (e.g. line length of 135?). i believe we should use the standard rustfmt settings, so at least for this directory i override the current global settings. i propose to remove the project-wide rustfmt configuration and reformat the project with the standard formatter.

If we keep this, I need to add a newline here

eng-wasm/derive/Cargo.toml
```diff
@@ -1,22 +1,24 @@
...
 license = "AGPL-3.0"
 description = "Enigma library for creating Secret Contracts"
 keywords = ["wasm", "webassembly", "blockchain", "sgx", "enigma"]
 categories = ["wasm"]
 
+[lib]
+proc-macro = true
+
 [dependencies]
-eng-wasm = { version = "0.1.3", path = "../" }
```
reuvenpo: Please note that this dependency is both unnecessary, and causes a future compatibility issue. I would like to create a dependency from `eng-wasm` to `eng-wasm-derive` so that we can reexport pub_interface from eng-wasm directly (and remove the requirement to depend on `eng-wasm-derive` directly. This can not happen without bumping the version of `eng-wasm` to 0.2 as it will not be compatible with older versions of `eng-wasm-derive` which depend on `eng-wasm`

eng-wasm/derive/Cargo.toml
```diff
@@ -1,22 +1,24 @@
...
+quote = "1.0"
+failure = "0.1"
 ethabi = "6.1"
 serde_json = "1.0"
 tiny-keccak = "1.4"
 
-[lib]
-proc-macro = true
+[dev-dependencies]
+syn = { version = "1.0", features = ["full", "extra-traits"] }
```
reuvenpo, line 25: I use the `extra-traits` feature for printing and comparisons of AST objects

eng-wasm/derive/src/into_ident.rs
```diff
@@ -0,0 +1,9 @@
+pub trait IntoIdent {
```
reuvenpo, line 1: I think this is fairly self-explanatory, but if not i can add a docstring. i'm surprised this is not implemented upstream, so we can file a pull request with this to https://github.com/dtolnay/syn

eng-wasm/derive/src/lib.rs
```diff
@@ -1,328 +1,49 @@
...
-        let return_params_number = match output{
-            syn::ReturnType::Type( _, box syn::Type::Path(n) ) => n.path.segments.len(),
-            _ => 0,
+use proc_macro::TokenStream;
+
+/// This is a proc_macro2 equivalent of syn::parse_macro_input.
+///
+/// I placed it here because [that's how macro_rules scoping works][macro-rules-scoping]
+/// and rust didn't like it when i put it in a module and exported it. If you figure out a
+/// better way to hide it, please file a PR.
```
reuvenpo, line 12: or comment here :)

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
...
+                }
+            }
+        }
+    }
+}
+
+fn generate_deploy_function(
+    signatures: &PubInterfaceSignatures,
+    deploy_func_name: &syn::Ident,
+) -> proc_macro2::TokenStream {
```
reuvenpo, line 182: I should reverse the order of arguments here

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
...
+const DEPLOY_FUNC_NAME: &str = "deploy";
+const DISPATCH_FUNC_NAME: &str = "dispatch";
+const FUNCTION_NAME_FUNC_NAME: &str = "function_name";
+const ARGS_FUNC_NAME: &str = "args";
+const CALL_FUNC_NAME: &str = "call";
+
+pub(crate) fn impl_pub_interface(item: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
+    //    let item = parse_macro_input2!(item as syn::ItemTrait);
+    //    println!("SYNTAX: {:#?}", item);
+    //    item.to_token_stream()
```
reuvenpo: I forgot to remove this

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
...
+
+                let return_value_count = match output_type {
+                    syn::ReturnType::Default => 0,
+                    // If the return value is a tuple, we count it like multiple return values.
+                    // This is the same thing that pwasm_abi does under
+                    // pwasm_abi/derive/src/item.rs :: fn into_signature
+                    // which flows back to
+                    // pwasm_abi/derive/src/lib.rs :: fn generate_eth_endpoint
+                    // which dictates how return values are serialised into the Sink.
+                    // This can be 0 is the return type is () which is correct.
```
reuvenpo: typo "if the return value is"

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
...
+    ImplMethodWithBadVisibility,
+}
+
+enum PubInterfaceInput {
+    ItemTrait(syn::ItemTrait),
+    ItemImpl(syn::ItemImpl),
+}
+
+/// The signatures collected while parsing the macro input
+#[allow(dead_code)]
```
reuvenpo: remove this annotation

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
...
+        .items
+        .into_iter()
+        // Filter out anything that isn't a trait method
+        .filter_map(get_trait_item_as_method)
+        // For each method in the definition, collect problems `Vec<syn::Error>`
+        .map(check_trait_method)
+        .partition::<Vec<_>, _>(|item| item.is_ok());
+
+    // Thanks to the .partition() above, we know that:
+    // all `signatures` are `Result::Ok(syn::Signature)`
```
reuvenpo: all `trait_methods`

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
...
+        .items
+        .into_iter()
+        // Filter out anything that isn't a public impl method
+        .filter_map(get_pub_impl_item_as_method)
+        // For each method in the definition, collect problems `Vec<syn::Error>`
+        .map(check_impl_method)
+        .partition::<Vec<_>, _>(|item| item.is_ok());
+
+    // Thanks to the .partition() above, we know that:
+    // all `signatures` are `Result::Ok(syn::Signature)`
```
reuvenpo: all `impl_methods`

reuvenpo:
I'll point out that this code relies on the order of evaluation of function arguments, which is unspecified:
https://www.reddit.com/r/rust/comments/cw0kyd/is_the_order_of_evaluation_of_function_arguments/
but is never going to change from left-to-right:
https://internals.rust-lang.org/t/rust-expression-order-of-evaluation/2605/29
and will be documented as such at some point:
https://github.com/rust-lang-nursery/reference/issues/248

EDIT: I removed this dependency anyways, it was fairly easy :)

eng-wasm/derive/src/lib.rs
```diff
@@ -1,328 +1,49 @@
-#![deny(unused_extern_crates)]
 #![feature(box_patterns)]
```
reuvenpo: This can be avoided:
https://github.com/dtolnay/syn/issues/701

eng-wasm/derive/src/into_ident.rs
moriaab, line 1: Nice addition!

eng-wasm/derive/src/lib.rs
```diff
@@ -1,333 +1,49 @@
-#![deny(unused_extern_crates)]
```
moriaab, line 1: Nice cleaning of attributes and dependencies :-)

eng-wasm/derive/src/into_ident.rs
reuvenpo: I just realised this is probably why it isn't made to be as easy to use as i made it:
https://docs.rs/quote/1.0.2/quote/macro.format_ident.html#panics

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
+use quote::{quote, ToTokens};
+use syn::spanned::Spanned;
+
+mod parse_signatures;
+
+use super::into_ident::IntoIdent;
+use parse_signatures::PubInterfaceSignatures;
+
+const DEFAULT_IMPLEMENTOR_NAME: &str = "Contract";
```
moriaab, line 9: Excellent addition of definition and work with string constants here!

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
...
+                            let mut stream = eng_wasm::eng_pwasm_abi::eth::Stream::new(args);
+                            <#implementor>::#method_name(#(stream.pop::<#input_types>().expect(#expectations),)*);
+                        }
+                    }),
+                    _ => Some(quote! {
+                        #method_name_as_string => {
+                            let mut stream = eng_wasm::eng_pwasm_abi::eth::Stream::new(args);
+                            let result = <#implementor>::#method_name(#(stream.pop::<#input_types>().expect(#expectations),)*);
+                            // 32 is the size of each argument in the serialised form
+                            // The Sink.drain_to() method might resize this array is any
```
reuvenpo: typo: The Sink.drain_to() method might resize this array **if** any

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
...
+
+use failure::Fail;
+
+use crate::into_ident::IntoIdent;
+use crate::reduce_mut::ReduceMut;
+
+/// Failures that can happen while parsing the macro input.
+///
+/// This type mostly exists to pass into syn::Error as an error message.
+#[derive(Fail, Debug)]
```
reuvenpo: use this instead of `Fail`:
https://crates.io/crates/parse-display

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
+//! This module defines the parsing of input for the `#[pub_interface]` macro
```
moriaab, line 1: parse_signatures.rs code sets high standards :) Excellent work!

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
...
+            input.advance_to(&forked_input);
+            pub_interface_input
+        })
+        // If none of the options worked, we tell the user that he gave us bad input.
+        .map_err(|_err| input.error(ParseError::BadInputItem))?;
+
+        let mut implementor: syn::Type;
+        let signatures = match pub_interface_input {
+            PubInterfaceInput::ItemTrait(item_trait) => {
+                let default_implementor_name = super::DEFAULT_IMPLEMENTOR_NAME.into_ident();
```
moriaab, line 121: Please pay attention: the fixed name `DEFAULT_IMPLEMENTOR_NAME` of the structure is a *defect* that you luckily revealed. Any name should be allowed. I will appreciate if you can add this fix as well.

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -0,0 +1,600 @@
...
+
+    // Thanks to the .partition() above, we know that:
+    // all `signatures` are `Result::Ok(syn::Signature)`
+    // and all `errors` are `Result::Err(Vec<syn::Error>)`
+    // so we can safely `.ok().unwrap()` and `.err().unwrap()` below.
+
+    if errors.is_empty() {
+        Ok(trait_methods
+            .into_iter()
+            .map(|res| res.ok().unwrap())
```
moriaab: If, as you realized, `ok()` is not needed here and in line 231, please remove it.

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
...
+}
+
+fn generate_deploy_function(
+    signatures: &PubInterfaceSignatures,
+    deploy_func_name: &syn::Ident,
+) -> proc_macro2::TokenStream {
+    if let Some(constructor_signature) = signatures
+        .signatures
+        .iter()
+        .find(|item| item.ident == CONSTRUCTOR_NAME)
```
moriaab, line 186: Please add one more check about the constructor: it should return nothing.

eng-wasm/derive/src/pub_interface.rs
```diff
@@ -0,0 +1,475 @@
...
+                let return_value_count = match output_type {
+                    syn::ReturnType::Default => 0,
+                    // If the return value is a tuple, we count it like multiple return values.
+                    // This is the same thing that pwasm_abi does under
+                    // pwasm_abi/derive/src/item.rs :: fn into_signature
+                    // which flows back to
+                    // pwasm_abi/derive/src/lib.rs :: fn generate_eth_endpoint
+                    // which dictates how return values are serialised into the Sink.
+                    // This can be 0 is the return type is () which is correct.
+                    syn::ReturnType::Type(_, box syn::Type::Tuple(return_tuple)) => return_tuple.elems.len(),
```
moriaab: Amazing catch and fix here!

eng-wasm/derive/src/pub_interface.rs
reuvenpo: I added this check, as well as a check that the constructor is marked as public when defined in an `impl`

eng-wasm/derive/src/pub_interface/parse_signatures.rs
reuvenpo: fixed :)

eng-wasm/derive/src/pub_interface/parse_signatures.rs
```diff
@@ -292,11 +363,35 @@ fn check_impl_method(
...
+
+    if errors.is_empty() {
         Ok(impl_method)
     } else {
         Err(errors)
-    };
+    }
+}
+
+/// This function checks if the function has a non-unit return type.
```
moriaab, line 403: Good work! It is so easy to miss the distinction between empty tuple and no return type.

moriaab (approved):
This PR improves `eng-wasm-derive` crate significantly in all aspects. Well done!

eng-wasm/derive/examples/trait-impl-rename.rs
```diff
@@ -0,0 +1,29 @@
...
+#[pub_interface(MyContractImplementation)]
+trait MyContract {
+    /// constructor
+    fn construct(_x: u32);
+
+    /// secret contract method
+    fn expand(input: u32) -> u64;
+}
+
+impl MyContract for MyContractImplementation {
```
moriaab, line 16: I would maybe add `impl MyContract` for another structure in addition, just to emphasize the point of the argument to `pub_interface`. But it is up to you.

PR #210, opened 2019-09-03
Fixed bug in encoding dynamic types, and upgraded dependencies 

AvishaiW:
upgraded: 

- eng-pwasm-abi 0.3.2
- eng-wasm 0.1.4
- eng-wasm-derive 0.1.4
- ethabi 8.0.1
- ethereum-types 0.6.0
- web3 0.8.0

Furthermore, this PR fixes the dynamic types that can be sent to the SC,  so It's available also to send:

- Vec<Vec<u8>>
- Vec<String>
and vectors of those types as well

### DIFF
--- a/enigma-principal/app/src/esgx/epoch_keeper_u.rs
+++ b/enigma-principal/app/src/esgx/epoch_keeper_u.rs
@@ -37,7 +37,7 @@ pub fn set_or_verify_worker_params(eid: sgx_enclave_id_t, worker_params: &InputW
     let mut retval: EnclaveReturn = EnclaveReturn::Success;
     println!("Evaluating nonce/seed based on EpochState: {:?}", epoch_state);
     let (nonce_in, seed_in) = match epoch_state.clone() {
-        Some(e) => (H256::from(e.nonce).0, H256::from(e.seed).0),
+        Some(e) => (e.nonce.into(), e.seed.into()),
         None => ([0; 32], [0; 32])
     };
     println!("Calling enclave set_worker_params with nonce/seed: {:?}/{:?}", nonce_in.to_vec().to_hex(), seed_in.to_vec().to_hex());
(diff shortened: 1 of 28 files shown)


examples/eng_wasm_contracts/encryption/Cargo.toml
```diff
@@ -3,10 +3,8 @@ name = "contract"
 version = "0.1.0"
 
 [dependencies]
-#eng-wasm = "0.1.3"
```
moriaab, line 6: Good catch.

enigma-principal/app/src/esgx/epoch_keeper_u.rs
```diff
@@ -37,7 +37,7 @@ pub fn set_or_verify_worker_params(eid: sgx_enclave_id_t, worker_params: &InputW
     let mut retval: EnclaveReturn = EnclaveReturn::Success;
     println!("Evaluating nonce/seed based on EpochState: {:?}", epoch_state);
     let (nonce_in, seed_in) = match epoch_state.clone() {
-        Some(e) => (H256::from(e.nonce).0, H256::from(e.seed).0),
```
moriaab, line 40: Nice simplification!

moriaab (approved):
Good work!

examples/eng_wasm_contracts/encryption/Cargo.toml
moriaab: Wait, bad catch :) The test of this contract will not pass, since the version with crypto service is not pushed yet. Please revert this.

moriaab: On the second thought, this worked by chance. The reason for a local dependency here is that the version of eng-wasm, which contains crypto service was not pushed yet to crates.io.  When you pushed version 0.1.4, you unintently pushed the version with crypto service, therefore this change is possible.

thaodt:
cool! ;)

PR #211, opened 2019-09-04
Store at the end of the ecall

AvishaiW:


enigma-core/enclave/src/lib.rs
```diff
@@ -352,6 +348,7 @@ unsafe fn ecall_execute_internal(pre_execution_data: &mut Vec<Box<[u8]>>, byteco
     // Signing: S(exeCodeHash, inputsHash, delta(X-1)Hash, deltaXHash, outputHash, usedGas, optionalEthereumData, Success)
     let used_gas = result.used_gas.to_be_bytes();
     let output_hash = encrypted_output.keccak256();
+    let delta_hash = save_enc_delta(db_ptr, &exec_res.state_delta)?;
```
moriaab: I think that there is no difference between saving state and saving delta in the aspect of this PR. Therefore, I suggest (1) to split creation of delta for the signing and saving of delta; (2) implement saving of delta and state in a manner that if one fails, the other fails as well.

moriaab (changes requested):
Please see my comment. I think this change is required, unless I miss something.

enigma-core/enclave/src/lib.rs
AvishaiW: I implemented this, however, there can be a scenario in which the delta is stored and the state isn't. in this situation, I can remove the delta from the DB but it seems to me like a bit of an overkill. what do you say?

enigma-core/enclave/src/lib.rs
```diff
@@ -266,12 +266,22 @@ fn decrypt_inputs(callable: &[u8], args: &[u8], inputs_key: &DhKey) -> Result<(V
...
     } else {
-        Ok(Hash256::default())
+        Hash256::default()
+    }
+}
+
+fn store_delta_and_state(db_ptr: *const RawPointer, delta: &Option<EncryptedPatch>, state: &ContractState) -> Result<(), EnclaveError> {
+    match delta {
+        Some(d) => {
+            enigma_runtime_t::ocalls_t::save_delta(db_ptr, d)?;
```
moriaab, line 281: Here I would suggest to minimize the probability of error between saving delta and state and move encryption of the state before saving delta.

enigma-core/enclave/src/lib.rs
moriaab: It depends on how hard the removing of delta form DB is. If the task is easy, it is better to implement it now. If it is hard of time consuming, maybe the best option is to address this in the comment and put this task in a backlog.

moriaab:
@AvishaiW 
I realized that the similar work should be done for KM node.

enigma-core/enclave/src/lib.rs
```diff
@@ -266,12 +266,22 @@ fn decrypt_inputs(callable: &[u8], args: &[u8], inputs_key: &DhKey) -> Result<(V
...
-        enigma_runtime_t::ocalls_t::save_delta(db_ptr, delta)?;
-        Ok(delta.keccak256_patch())
+        delta.keccak256_patch()
     } else {
-        Ok(Hash256::default())
+        Hash256::default()
+    }
+}
+
+fn store_delta_and_state(db_ptr: *const RawPointer, delta: &Option<EncryptedPatch>, state: &ContractState) -> Result<(), EnclaveError> {
```
reuvenpo: Sorry for barging into the code review.

This function must be marked as `unsafe`, since passing in an invalid pointer as the `dp_ptr` argument will trigger UB. (Anything from a segfault to invalid memory access to full process corruption)

enigma-core/enclave/src/lib.rs
AvishaiW: done

enigma-core/enclave/src/lib.rs
AvishaiW: done

AvishaiW:
@moriaab currently we have no calls to the DB in the KM

enigma-core/enclave/src/lib.rs
AvishaiW: this case will corrupt the DB forever, therefore, I added a todo for this, and it will be implemented once we add the option for reverting the state.

moriaab (approved):
Will "todo" issue of reverting stored delta be addressed as part of revert state task? 
If so, it is important to update `store_delta_and_state` function accordingly. 
If not, let's talk when it should be implemented.

AvishaiW:
@moriaab yes

PR #212, opened 2019-09-04
retrieve the SC addresses from the getAllSCAddresses getter instead of two calls to the contract

AvishaiW:


enigma-tools-u/src/web3_utils/enigma_contract.rs
```diff
@@ -180,4 +183,16 @@ impl ContractQueries for EnigmaContract {
...
+    #[logfn(INFO)]
+    fn get_all_secret_contract_addresses(&self) -> Result<Vec<ContractAddress>, Error> {
+        let addrs: Vec<H256> =
+            match self.w3_contract.query("getAllSecretContractAddresses", (),self.account, Options::default(), None).wait() {
+                Ok(addrs) => addrs,
+                Err(e) => {
+                    return Err(errors::Web3Error { message: format!("Unable to query getAllSecretContractAddresses: {:?}", e) }.into())
+                }
+            };
+        Ok(addrs.into_iter().map(|a|ContractAddress::from(a.0)).collect())
```
moriaab: It is very minor, but I think that this line may move to the line 191, then there is no need to defile `addrs` variable.

moriaab (approved):

PR #213, opened 2019-09-08
updating state during computations

AvishaiW:
This PR adds calls to building the state when a computation is executed in a scenario in which the deltas were updated after the epoch has changed (until now, we built the state for all contracts addressed to the worker after the ptt)- this can be possible in a case that the previous worker working on a contract was assigned a task which he finished only after the ptt process, leading the new worker to execute a new task on the wrong state.


enigma-core/app/src/db/dal.rs
```diff
@@ -15,6 +15,8 @@ pub struct DB {
     // the DB needs to store the options for creating new
     // cf's that would be able to imitate the DB behaviour
     pub options: Options,
+    // keeps track of the status of the state
+    is_stateful: bool,
```
moriaab: I think that this name `is_stateful` and the comment do not express the meaning of the attribute.

enigma-core/app/src/db/dal.rs
```diff
@@ -53,9 +55,23 @@ impl DB {
         let cf_list_burrowed = cf_list.iter().map(String::as_str).collect::<Vec<&str>>();
         let database = rocks_db::open_cf(&options, &location, &cf_list_burrowed[..])?;
         let location = location.as_ref().to_path_buf();
-        let db_par = DB { location, database, options };
+        // the is_stateful is initialized to true since it won't be necessary to build
```
moriaab: This comment is very unclear to me. Let's discuss it.

enigma-core/app/src/networking/ipc_listener.rs
```diff
@@ -281,6 +281,8 @@ pub(self) mod handling {
             let delta = IpcStatusResult { address, key, status };
             errors.push(delta);
         }
+        // since a new delta was added the state is no longer updated
+        db.update_state_status(false);
```
moriaab, line 287: I didn't get into details, so maybe this comment is wrong, but shouldn't the same be done for `update_new_contract_on_deployment`?

enigma-core/app/src/db/dal.rs
AvishaiW: will change it. `state_updated` is sufficient?

enigma-core/app/src/networking/ipc_listener.rs
AvishaiW: at first, I thought we won't need it, however, now that I'm thinking of it, you're right. I'll add this as well.

AvishaiW: We don't need it, since every contract comes with a delta as well.

AvishaiW: you're right! we do need this. I added it.

enigma-core/app/src/networking/ipc_listener.rs
```diff
@@ -314,6 +318,8 @@ pub(self) mod handling {
         let msg = response.response.from_hex()?;
         km_u::ptt_res(eid, &msg)?;
         let res = km_u::ptt_build_state(db, eid)?;
+        // update state status
```
moriaab: This comment is redundant, since it repeats the function name. I would add here a more informative comment like you did  in case of `db.update_state_status(false)`

moriaab (approved):
To my opinion, this is a midtime solution. One indicator `state_updated` for all contracts is OK at this moment as fast and easy solution. **The long-term solution and its timing should be discussed**.

Please add description for this PR, it is not trivial one.

PR #215, opened 2019-09-08
Added the option to add the config in env variables

AvishaiW:


moriaab (approved):
`envy` is a nice finding, it saves boring work!
Please *add test* to the case of initialization with the environment variables.

moriaab (approved):
LGTM. Please merge this.

PR #221, opened 2019-09-18
EP 688 sign gas limit

reuvenpo:
This PR adds a few minor fixes, and the final commit adds the gas limit to the information that is used when signing task success/failure and deployment success/failure. This addition will allow enigma contract (and users) to know that the worker has not (maliciously) given the enclave a gas limit which is very low, which would always cause task failures.

another notable change is that enigma-types is now no-std by default instead of std by default. This is done as a consequence of the fact that cargo features are almost exclusively additive, which caused unused crates to be downloaded when compiling it for sgx. Not the default features are as minimal as possible, and any additional runtime functionality (std or sgx) needs to be explicitly specified.


enigma-core/enclave/src/lib.rs
```diff
@@ -126,9 +126,9 @@ pub unsafe extern "C" fn ecall_execute(bytecode: *const u8, bytecode_len: usize,
                            *gas_limit,
                            db_ptr,
                            result);
-    if let Err(e) = internal_result.clone() {
```
moriaab, line 131: Nice!

enigma-core/enclave/src/lib.rs
```diff
@@ -396,19 +399,17 @@ unsafe fn ecall_deploy_internal(pre_execution_data: &mut Vec<Box<[u8]>>, bytecod
...
                         result)?;
 
-//    let exe_code = &exec_res.result[..];
-//    *output_ptr = ocalls_t::save_to_untrusted_memory(&exe_code)?;
-
-    // Signing: S(inputsHash, exeCodeHash, delta0Hash, usedGas, optionalEthereumData, Success)
+    // Signing: S(inputsHash, exeCodeHash, delta0Hash, gasLimit, usedGas, optionalEthereumData, Success)
     let used_gas = result.used_gas.to_be_bytes();
     let (ethereum_payload, ethereum_address) = create_eth_data_to_sign(exec_res.ethereum_bridge);
-    let to_sign = [
```
moriaab, line 411: It is much cleaner now.

moriaab (approved):
The gas limit signing part looks good.
Thank you for improving other parts of code as well :-)

.drone.yml
```diff
@@ -70,9 +70,10 @@ steps:
       - . /opt/sgxsdk/environment && . /root/.cargo/env
       - cd enigma-types
       - RUSTFLAGS=-Awarnings cargo test
-      - RUSTFLAGS=-Awarnings cargo build --no-default-features
-      - RUSTFLAGS=-Awarnings cargo build --no-default-features --features=sgx
-      - RUSTFLAGS=-Awarnings cargo build --no-default-features --features=alloc
+      - RUSTFLAGS=-Awarnings cargo check
```
moriaab, line 81: Nice change from `build` to `check`

reuvenpo:
https://github.com/enigmampc/enigma-core/pull/221

PR #229, opened 2019-09-26
Changes sealing policy to MRSIGNER

AvishaiW:
This PR changes the sealing policy in order to avoid malfunctions during upgrades. this way we would be able to continue running with the same keypair from the previous version of the enclave


moriaab (approved):
LGTM

PR #231, opened 2019-10-03
Retry Attestation calls at registration

AvishaiW:
This PR adds a retry loop for getting the attestation report from the proxy server. we can get the number of retries from the user or run 10 by default

### DIFF
--- a/enigma-principal/app/src/boot_network/principal_manager.rs
+++ b/enigma-principal/app/src/boot_network/principal_manager.rs
@@ -415,5 +415,6 @@ mod test {
         let config = PrincipalConfig::load_config("this is not a path").unwrap();
         assert_eq!(config.polling_interval, 1);
         assert_eq!(config.http_port, 3040);
+        assert_eq!(config.attestation_retries, 11);
     }
 }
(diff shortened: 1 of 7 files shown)


lacabra:
Thanks @AvishaiW for working on this, very much appreciated! 🙏

enigma-principal/app/src/boot_network/principal_manager.rs
```diff
@@ -408,6 +409,7 @@ mod test {
         env::set_var("MAX_EPOCHS","10");
         env::set_var("SPID", "B0335FD3BC1CCA8F804EB98A6420592D");
         env::set_var("ATTESTATION_SERVICE_URL", "https://sgx.enigma.co/api");
+        env::set_var("ATTESTATION_RETRIES", "11");
```
moriaab, line 412: I understand that it is only testing parameter, but just out of curiosity, why the value is different from the value in core?

AvishaiW: I wanted to make sure that it's getting the value from the environment variable and not as the default value when initializing the KM without anything.

moriaab (approved):
Excellent!
Only one comment, which you may implement later. There are multiple places with definition of the constant for the number of retries in production: `cli.rs`, `mod.rs`, `principal_test_config.json`, `service.rs`. Please define it in one place and reference it.

PR #232, opened 2019-10-07
MAINT: builds match branches across repos

lacabra:
The CI now checks for matching branches across the other two repos, and conditionally builds against those other branches if a match is found. The logic is as follows:

1. If branch equals `master`, then use `latest` docker images
2. Else, if feature branch does NOT have a match in other repos, then use `develop` docker images
3. Else, if there is a match in only one other repo, then build a local image using that remote branch, AND reuse the existing `develop` for the repo there is no match (by tagging it as the one that will be used)
4. Else (there is a match on both repos), then build both local images.

Because of concurrency challenges derived from the fact that now three different repos `core`, `p2p` and `contract` handle the CI builds in the same host and they share docker resources, the images need to be separated between builds to avoid cross-contamination if there is a bogus image. Thus all images are tagged with `repo_build`, and deleted at the end of the build.

Analogous to:
- enigmampc/enigma-contract#159
- enigmampc/enigma-p2p#236

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -99,7 +99,7 @@ steps:
           export TAG=latest;
           docker-compose pull
         else
-          if ! (( $MATCHING_BRANCH_P2P + $MATCHING_BRANCH_CONTRACT )); then
+          if [ $MATCHING_BRANCH_P2P -eq 0 ] &&  [ $MATCHING_BRANCH_CONTRACT -eq 0 ]; then
             export TAG=develop;
             sed -i "s/DOCKER_TAG=latest/DOCKER_TAG=develop/" .env;
             docker-compose pull
@@ -127,7 +127,10 @@ steps:
         docker build --build-arg GIT_BRANCH_CORE=$DRONE_BRANCH --build-arg SGX_MODE=HW -t enigmampc/enigma_core_hw:$TAG --no-cache . &&
         docker build -f Dockerfile.km --build-arg GIT_BRANCH_CORE=$DRONE_BRANCH --build-arg SGX_MODE=HW -t enigmampc/enigma_km_hw:$TAG --no-cache . &&
         cd ..
-      - export NODES=3 && docker-compose -f docker-compose.yml -f docker-compose.hw.yml -f docker-compose.test.yml up --scale core=$NODES --scale p2p=$NODES --exit-code-from client
+      - export NODES=3 
+      - docker-compose -f docker-compose.yml -f docker-compose.hw.yml -f docker-compose.test.yml up --scale core=$NODES --scale p2p=$NODES --exit-code-from client && export RESULT=$? || export RESULT=$?
+      - if [[ "${DRONE_BRANCH}" == "$TAG" ]]; then docker-compose down -v --rmi all; else docker-compose down -v; fi
+      - if [ $RESULT -ne 0 ]; exit 1; fi
     privileged: true
     volumes:
       - name: sock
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -85,29 +85,63 @@ steps:
 
   - name: integration
     image: enigmampc/docker-client
+    privileged: true
     depends_on:
       - core
       - principal
+    volumes:
+      - name: sock
+        path: /var/run/docker.sock
```
AvishaiW, line 94: why is necessary? what changed that it is needed in the integration tests as well?

AvishaiW (approved):
Great!

.drone.yml
```diff
@@ -85,29 +85,63 @@ steps:
...
+                cd enigma-contract && docker build --build-arg GIT_BRANCH_CONTRACT=${DRONE_BRANCH} -t enigmampc/enigma_contract:$DOCKER_TAG --no-cache . && cd ..
+                docker pull enigmampc/enigma_p2p:develop
+                docker tag enigmampc/enigma_p2p:develop enigmampc/enigma_p2p:$DOCKER_TAG
+              else
+                cd enigma-p2p && docker build --build-arg GIT_BRANCH_P2P=${DRONE_BRANCH} -t enigmampc/enigma_p2p:$DOCKER_TAG --no-cache . && cd ..
+                cd enigma-contract && docker build --build-arg GIT_BRANCH_CONTRACT=${DRONE_BRANCH} -t enigmampc/enigma_contract:$DOCKER_TAG --no-cache . && cd ..
+              fi
+            fi
+          fi
+        fi
```
reuvenpo: This should be more maintainable:
```bash
export MATCHING_BRANCH_P2P="$(git ls-remote --heads https://github.com/enigmampc/enigma-p2p.git ${DRONE_BRANCH} | wc -l)"
export MATCHING_BRANCH_CONTRACT="$(git ls-remote --heads https://github.com/enigmampc/enigma-contract.git ${DRONE_BRANCH} | wc -l)"
export DOCKER_TAG=core_${DRONE_BUILD_NUMBER}
sed -i "s/DOCKER_TAG=latest/DOCKER_TAG=${DOCKER_TAG}/" .env;

declare -a PROJECTS=(p2p contract)

declare -A DOCKER_IMAGES=(
    [p2p]=enigma_p2p
    [contract]=enigma_contract
)

declare -A GIT_BRANCH_ARG=(
    [p2p]=GIT_BRANCH_P2P
    [contract]=GIT_BRANCH_CONTRACT
)

declare -A PROJECT_BRANCH_FOUND=(
    [p2p]=MATCHING_BRANCH_P2P
    [contract]=MATCHING_BRANCH_CONTRACT
)

declare -A PROJECT_DIRECTORY=(
    [p2p]=enigma-p2p
    [contract]=enigma-contract
)

for project in "${PROJECTS}"; do
    DOCKER_IMAGE="enigmampc/${DOCKER_IMAGES[$project]}"
    if [[ "$DRONE_BRANCH" == "master" ]]; then
        docker pull "$DOCKER_IMAGE:latest"
        docker tag "$DOCKER_IMAGE:latest" "$DOCKER_IMAGE:$DOCKER_TAG"
    # The ! mark here is "indirect expansion"
    elif [ "${!PROJECT_BRANCH_FOUND[$project]}" -eq 0 ]; then
        docker pull "$DOCKER_IMAGE:develop"
        docker tag "$DOCKER_IMAGE:develop" "$DOCKER_IMAGE:$DOCKER_TAG"
    elif [ "${!PROJECT_BRANCH_FOUND[$project]}" -eq 1 ]; then
        pushd "${PROJECT_DIRECTORY[$project]}" && docker build --build-arg "${GIT_BRANCH_ARG[$project]}=${DRONE_BRANCH}" -t "$DOCKER_IMAGE:$DOCKER_TAG" --no-cache . && popd
    fi
done
```

reuvenpo: Oh... i just realized you can remove the indirection trick in the `elif` clause if you expand the `MATCHING_BRANCH_*` parts in the definition of `PROJECT_BRANCH_FOUND` (just prefix a `$`), but i'll keep it there so you can see that it's possible if you need it :+1:

.drone.yml
lacabra: Nothing has changed, I am only re-ordering lines. If you take a look at the file before these changes, these lines were already present in 105-107. Because the container needs to access docker on the host, it needs access to the socket.

.drone.yml
lacabra: Thanks @reuvenpo for the suggestion. After much trial and error, this has been added in 5917a2f

PR #235, opened 2019-10-10
Adding simple factorization contract + test

moriaab:


enigma-core/app/src/wasm_u/wasm.rs
```diff
@@ -374,6 +374,24 @@ mod tests {
...
+            &mut db,
+            "../../examples/eng_wasm_contracts/factorization",
+            generate_contract_address(),
+            "construct()",
+            &[],
+            "find_number_of_prime_factors(uint32)",
+            &[Token::Uint(199998.into())]
+        );
+        let encoded_output = symmetric::decrypt(&result.output, &shared_key).unwrap();
+        let decoded_output = &(ethabi::decode(&[ethabi::ParamType::Uint(64)], &encoded_output).unwrap())[0];
```
reuvenpo: You can remove one pair of parens:
```rust
let decoded_output = &ethabi::decode(&[ethabi::ParamType::Uint(64)], &encoded_output).unwrap()[0];
```

enigma-core/app/src/wasm_u/wasm.rs
```diff
@@ -374,6 +374,24 @@ mod tests {
...
+        assert_eq!(decoded_output.clone().to_uint().unwrap().as_u64(), 5);
```
reuvenpo, line 393: why clone?

examples/eng_wasm_contracts/factorization/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+eng-wasm-derive = "0.1"
+
+[lib]
+crate-type = ["cdylib"]
+
+[profile.release]
+panic = "abort"
+lto = true
+opt-level = "z"
+#overflow-checks = true # Think about this, it will increase size and performance but I think it might be crucial
```
reuvenpo: you can remove this comment i think (although that is a really good question!)

examples/eng_wasm_contracts/factorization/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use eng_wasm_derive::pub_interface;
+
+struct S;
+
+#[pub_interface]
+impl S {
+    pub fn find_number_of_prime_factors(mut number: u64) -> u64{
+        let mut result: u64 = 0;
+        let factor_bound = ((number as f64).sqrt().round() + 1.0) as u64;
```
reuvenpo: I think that `.ceil()` is better here. I'm not sure though.
```rust
(number as f64).sqrt().round() + 1.0
```

examples/eng_wasm_contracts/factorization/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        // The maximal number of 2s that divide `number`
+        while number & 1 == 0 {
+            result += 1;
+            number = number.clone() >> 1;
+        }
```
reuvenpo: https://doc.rust-lang.org/std/primitive.u64.html#method.trailing_zeros
```rust
result = number.trailing_zeroes();
number >>= result;
```
The method trailing_zeroes is an inline function defined with an intrinsic. this means it should generate less code (but we should verify this. Do we have a good WASM code explorer? Maybe Godbolt or ghidra/radare?)

reuvenpo: also note that in your code, the clone is redundant (but should be compiled away anyway)

examples/eng_wasm_contracts/factorization/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        for i in (3..factor_bound).step_by(2) {
+            while number % i == 0 {
+                result += 1;
+                number = number.clone() / i;
```
reuvenpo: the clone is redundant:
```rust
number /= i;
```

reuvenpo (commented):
I'll add that i don't know our testing mechanism well enough to tell if your test in wasm.rs can be improved significantly. On the surface level it looks good and i can tell what it's doing. I would throw it into a loop which runs the same contract multiple times with several numbers, and checks that the result is correct for all of them.

enigma-core/app/src/wasm_u/wasm.rs
moriaab: Done

examples/eng_wasm_contracts/factorization/Cargo.toml
moriaab: Removed and created a future task for you to investigate this :)

examples/eng_wasm_contracts/factorization/src/lib.rs
moriaab: Done

examples/eng_wasm_contracts/factorization/src/lib.rs
moriaab: COOL! I don't have time to check whether it generates less code or not, but it is more elegant, so I choose your approach.

examples/eng_wasm_contracts/factorization/src/lib.rs
moriaab: Done

enigma-core/app/src/wasm_u/wasm.rs
moriaab: This is what we get without the clone:
move occurs because `*decoded_output` has type `km_u::tests::ethabi::Token`, which does not implement the `Copy` trait

reuvenpo (approved):

PR #236, opened 2019-10-16
MAINT: bumping to 0.3.0

moriaab:
Bumping crates to 0.3.0 towards creating of the new release


reuvenpo:
As discussed, we want to start maintaining a changelog file, but because we need this version bump right now, we'll do it in another PR.
חג שמח מוריה :)

reuvenpo (approved):

PR #238, opened 2019-10-18
Adding changelogs for 0.3.0

moriaab:
@reuvenpo 
I don't have a strong opinion on question: is this the best form to separate the changelog files for `enclave` and `app`? Some of the changes are relevant for the both.
In addition, please review:
* The format of the changelogs
* English
* Belonging of the items to the _"breaking changes"_
Thank you!


enigma-core/app/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+### Version 0.3.0
+
+#### Breaking Changes
+
+* Upgrading to use Rust SGX SDK **v1.0.9**
+* Adding retries for getting the attestation report from the proxy server. The number of retries may be obtained as an input, otherwise the default number is used.
```
reuvenpo: I think that this isn't a breaking change..?
Is there a reason to think otherwise?

enigma-core/enclave/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+### Version 0.3.0
+
+#### Breaking Changes
+
+* Upgrading to use Rust SGX SDK **v1.0.9**
+* Adding ability to revert state
```
reuvenpo: This is only a breaking change if it affects the protocol

enigma-core/enclave/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+
+* Upgrading to use Rust SGX SDK **v1.0.9**
+* Adding ability to revert state
+* `enigma-types` is now `no-std` by default. The motivation is that cargo features are almost exclusively additive, which caused unused crates to be downloaded when compiling it for `sgx`.
+* `gas limit` is now part of the information that is used when signing task success/failure. This allows to verify that the worker cannot (maliciously) cause task failure by giving the enclave a low `gas limit`.
+* changing sealing policy to `SGX_KEYPOLICY_MRSIGNER`.
+
+#### Bug Fixes
+
+* The length field in serialization of multiple values for hashing is now always a `u64` independent of the platform.
```
reuvenpo: "... independent of the platform running the system (32 vs 64 bit)"

enigma-core/enclave/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+* Building the state when a computation is executed in a scenario in which the deltas were updated after the epoch has changed.
```
reuvenpo: I'm not quite sure what you mean here. Did you mean:

When a computation is executed after the deltas are updated following an epoch change, *the (contract?) state is rebuilt*

enigma-core/enclave/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+* Moving storing updated state and delta to the very end of a computation, which prevents possible inconsistency.
```
reuvenpo: storing of updated state and delta is moved to the very end of a computation ...

enigma-principal/app/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+### Version 0.3.0
+
+#### Breaking Changes
+
+* Upgrading to use Rust SGX SDK **v1.0.9**
+* Adding retries for getting the attestation report from the proxy server. The number of retries may be obtained as an input, otherwise the default number is used.
+* `enigma-types` is now `no-std` by default. The motivation is that cargo features are almost exclusively additive, which caused unused crates to be downloaded when compiling it for `sgx`.
+* Added the ability to add the configuration through environment variables
```
reuvenpo: This is technically a breaking change, but it can be argued that it  isn't. We can keep it here though.

reuvenpo (changes requested):
I think that the format of the changelog is good :)
The points which i mentioned don't belong in "breaking changes" can go in a section before that calles "changes" or "improvements". "improvements is opinionated though.
I think it's ok to combine these files into two files, one at enigma-core and one at enigma-proncipal. the app and enclave parts are always used together (as far as i know) so there isn't much point in talking about them separately.

enigma-core/app/CHANGELOG.md
moriaab: In my opinion the reason is that there is a new input now, which means API change.

enigma-core/enclave/CHANGELOG.md
moriaab: Yes, it affects the protocol: after the computation is done, the worker updates its state, but if the commit receipt fails, then the p2p part instructs the core to revert the state.

enigma-core/app/CHANGELOG.md
moriaab: After discussion with @AvishaiW, we decided that it is not a braking change because there is a default value for the new input. Thanks!

enigma-principal/app/CHANGELOG.md
moriaab: After your explanation, I prefer to treat it as a breaking change.

enigma-core/enclave/CHANGELOG.md
moriaab: Now the state is rebuilt if it was updated by others after PTT.

enigma-core/enclave/CHANGELOG.md
moriaab: Done

enigma-core/enclave/CHANGELOG.md
moriaab: Done

reuvenpo (approved):

PR #239, opened 2019-10-22
BUG: update reference smart -> secret contract

lacabra:
Error messages refer to "smart contract" when it should be "secret contract"


moriaab (approved):
Thanks!

PR #240, opened 2019-10-22
BLD: secret contract to test revert state

lacabra:
Modified `contract_with_eth_calls` to include `sum_and_call()` and `get_last_sum()` functions to test the functionality to revert the state in the network.

You will notice that the following lines feel hacky:
```
        let mut eth_addr_str_0x: String = "0x".to_string();
        let eth_addr_str: String = eth_addr.to_hex();
        eth_addr_str_0x.push_str(&eth_addr_str);
```
this is because if I omit them, and just pass the result of `eth_addr.to_hex()` to `EthContract::new()`, that function will cut the first two characters expecting `0x` but without checking, and will shift all characters apending `00` at the end, and effectively changing the Ethereum address that is being passed to the function. In my opinion, this is a bug of the constructor of `EthContract::new()`, and I believe it should be fixed. In the interim, this solution works.

This PR is missing the corresponding unit tests. I would ask @AvishaiW or @reuvenpo to please add them, if necessary. Thank you. 

This secret contract is being used in enigmampc/enigma-contract#166


reuvenpo:
@lacabra i agree that this is a bug of `EthContract::new()`. Really, why should it even require the 0x, but if it does, it should be validated, and panic or return a Result type. even better is if it accepts an address _type_ as input, rather than a String.
In any case, i'm pretty sure you can just write:
```rust
let mut eth_addr_str_0x: String = "0x".to_string() + eth_addr.to_hex();
```
plus or minus a few `&`'s if it complains about ownership.

examples/eng_wasm_contracts/contract_with_eth_calls/Test.json
```diff
@@ -1 +1,20 @@
-[{"constant":false,"inputs":[{"name":"","type":"uint256"},{"name":"","type":"address[]"}],"name":"getBryn","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"}]
+[
+	{
+		"constant":false,
+		"inputs":[{"name":"","type":"uint256"},{"name":"","type":"address[]"}],
```
reuvenpo: (I'm super picky, i know)
please add a bit of spacing after colons `:` and commas `,`
same on line 13

examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs
```diff
@@ -52,5 +58,23 @@ impl ContractInterface for Contract {
         let key = "code";
         eprint!("{}", a);
         write_state!(key => &a);
+        let sum: u64 = 12;
```
reuvenpo, line 61: why 12? is this value special?
if not, better set it to 0 or something

examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs
```diff
@@ -52,5 +58,23 @@ impl ContractInterface for Contract {
...
+        let c = EthContract::new(&eth_addr_str_0x);
+        c.record(sum.into());
+        sum.into()
+    }
+
+    fn get_last_sum() -> U256 {
+        let sum: u64 = read_state!("sum").unwrap_or_default();
+        sum.into()
+    }
+}
```
reuvenpo: Please add a newline at the end of files. Git complains if you omit it (the red arrow here)

reuvenpo (changes requested):

examples/eng_wasm_contracts/contract_with_eth_calls/Test.json
lacabra: fixed in 610284d

examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs
lacabra: why 0 then?
funny how you say `better set it to 0 or something`, then I set it to **something**, and that something is 12. Seriously, I don't care the value, I just want it to be something different than null and 0. The tests are plagued with random combination of values, this is no different.

Just from the same testfile:
- https://github.com/enigmampc/enigma-core/blob/c467f5f83253965330d39f9500718220fb97fd6c/examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs#L31
- https://github.com/enigmampc/enigma-core/blob/c467f5f83253965330d39f9500718220fb97fd6c/examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs#L51

So the answer is that is a discrete value that I can easily verify at the other end.

examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs
lacabra: fixed in 6612454

lacabra:
@reuvenpo: I agree with your assessment that this is a bug. 

I am not going to spend time here changing the syntax of those assignments since it works already, and it's going to change once the bug is fixed. This will have to be changed anyway.

I have addressed all of your other comments. Thanks for your review.

examples/eng_wasm_contracts/contract_with_eth_calls/src/lib.rs
reuvenpo: That's fair, i can understand that.

reuvenpo (approved):

PR #241, opened 2019-10-22
Updated to MRSIGNER

AvishaiW:
Missed this key policy call- saw it's defined as MRENCLAVE [here](https://github.com/baidu/rust-sgx-sdk/blob/v1.0.9/sgx_tseal/src/seal.rs#L240) so changed to MRSIGNER explicitly


moriaab (approved):
OMG! This is sealing of state keys and epoch info! Good catch.

Issue #230, opened 2019-09-26
A couple of issues with crypto_wasm::encrypt


Cashmaney:
@moriaab
A couple of issues I noticed when looking over the code @ [pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8>](https://github.com/enigmampc/enigma-core/blob/e0bfb8fd7260f6c929db7d06ec7e01dd067d87ed/eng-wasm/src/crypto_wasm.rs#L19)

- The encrypted buffer will always be the number of encryption blocks that the message fits into. So, instead of using PAYLOAD_SIZE you can determine the encrypted payload size by using message.len() + BLOCK_SIZE - (message.len() % BLOCK_SIZE).

It's also a good idea to check for upper/lower bounds for message.len() when making this change to avoid over allocating in case of malformed or malicious input. 

  After removing it you can remove: 

```rust
let mut end = 0;
for i in (0..length).rev() {
    if payload[i] != 0 {
        // Range selectors exclude the upper bound
        end = i + 1;
        break;
    }
}
payload[0..end].to_vec()
```

*By the way, the snippet above will probably fail if the last byte of the encryption is 0.

- Currently there's a possible buffer overflow in the unsafe function 
[`unsafe { external::encrypt(message.as_ptr(), message.len() as u32, key.as_ptr(), payload.as_mut_ptr()) };`](https://github.com/enigmampc/enigma-core/blob/e0bfb8fd7260f6c929db7d06ec7e01dd067d87ed/eng-wasm/src/crypto_wasm.rs#L27)

if message.len() is greater than PAYLOAD_SIZE, since from what I can understand you're copying the returned buffer from enigma_crypto::symetric::encrypt in an unsafe manner.

Solving the first issue will solve this as well, but it's still something to keep an eye on.

PR #242, opened 2019-10-24
Ep 822 finish crypto service

moriaab:
Solving of issue https://github.com/enigmampc/enigma-core/issues/230 + small fixes regarding crypto service

### DIFF
--- a/eng-wasm/src/crypto_wasm.rs
+++ b/eng-wasm/src/crypto_wasm.rs
@@ -4,10 +4,13 @@ use eng_pwasm_abi::types::U256;
 use rand_wasm::Rand;
 
 const SYMMETRIC_KEY_SIZE: usize = 32;
-const IV_SIZE: usize = 96 / 8;
-const TAG_SIZE: usize = 16;
 
 pub type SymmetricKey = [u8; SYMMETRIC_KEY_SIZE];
+
+fn extra_size_for_cypher() -> usize {
+    unsafe {(external::get_iv_size() + external:: get_tag_size()) as usize }
+}
+
 
 pub fn generate_key() -> SymmetricKey {
     let key_int: U256 = Rand::gen();
@@ -16,14 +19,14 @@ pub fn generate_key() -> SymmetricKey {
 }
 
 pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    let length = message.len() + IV_SIZE + TAG_SIZE;
+    let length = message.len() + extra_size_for_cypher();
     let mut payload = vec![0u8; length];
     unsafe { external::encrypt(message.as_ptr(), message.len() as u32, key.as_ptr(), payload.as_mut_ptr()) };
     payload.to_vec()
 }
 
 pub fn decrypt(cipheriv: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    let length: usize = cipheriv.len() - IV_SIZE - TAG_SIZE;
+    let length = cipheriv.len() - extra_size_for_cypher();
     let mut payload = vec![0u8; length];
     unsafe { external::decrypt(cipheriv.as_ptr(), cipheriv.len() as u32, key.as_ptr(), payload.as_mut_ptr()) };
     payload.to_vec()
--- a/eng-wasm/src/crypto_wasm.rs
+++ b/eng-wasm/src/crypto_wasm.rs
@@ -9,10 +9,13 @@ use rand_wasm::Rand;
 
 const SYMMETRIC_KEY_SIZE: usize = 256 / 8;
 pub type SymmetricKey = [u8; SYMMETRIC_KEY_SIZE];
+const AES_256_GCM_TAG_SIZE: usize = 16;
+const AES_256_GCM_IV_SIZE: usize = 96 / 8;
+
 
 /// The extra length (IV and authentication tag) of the encryption result
 fn extra_size_for_encrypted_text() -> usize {
-    unsafe {(external::get_iv_size() + external::get_tag_size()) as usize }
+    unsafe {AES_256_GCM_IV_SIZE + AES_256_GCM_TAG_SIZE}
 }
 
 
--- a/eng-wasm/src/crypto_wasm.rs
+++ b/eng-wasm/src/crypto_wasm.rs
@@ -5,9 +5,8 @@ use rand_wasm::Rand;
 
 const SYMMETRIC_KEY_SIZE: usize = 32;
 const IV_SIZE: usize = 96 / 8;
-const PAYLOAD_SIZE: usize = 1024;
+const TAG_SIZE: usize = 16;
 
-pub type IV = [u8; IV_SIZE];
 pub type SymmetricKey = [u8; SYMMETRIC_KEY_SIZE];
 
 pub fn generate_key() -> SymmetricKey {
@@ -17,41 +16,15 @@ pub fn generate_key() -> SymmetricKey {
 }
 
 pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    // TODO: Is this really needed? Dynamically sized buffers don't seem to work.
-    // TODO: Is it possible to estimate the encrypted payload size based on the plaintext message?
-    let length = PAYLOAD_SIZE;
-    let mut payload = Vec::with_capacity(length);
-    for _ in 0..length {
-        payload.push(0);
-    }
+    let length = message.len() + IV_SIZE + TAG_SIZE;
+    let mut payload = vec![0u8; length];
     unsafe { external::encrypt(message.as_ptr(), message.len() as u32, key.as_ptr(), payload.as_mut_ptr()) };
-    // Finding the end of trailing zeros
-    let mut end = 0;
-    for i in (0..length).rev() {
-        if payload[i] != 0 {
-            // Range selectors exclude the upper bound
-            end = i + 1;
-            break;
-        }
-    }
-    payload[0..end].to_vec()
+    payload.to_vec()
 }
 
 pub fn decrypt(cipheriv: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    // Assuming that plaintext messages cannot be shorter that their encrypted cipher
-    // TODO: Some unnecessary bytes can be subtracted from the buffer, at least the IV size.
-    let length: usize = cipheriv.len();
-    let mut payload = Vec::with_capacity(length);
-    for _ in 0..length {
-        payload.push(0);
-    }
+    let length: usize = cipheriv.len() - IV_SIZE - TAG_SIZE;
+    let mut payload = vec![0u8; length];
     unsafe { external::decrypt(cipheriv.as_ptr(), cipheriv.len() as u32, key.as_ptr(), payload.as_mut_ptr()) };
-    let mut end = 0;
-    for i in (0..length).rev() {
-        if payload[i] != 0 {
-            end = i + 1;
-            break;
-        }
-    }
-    payload[0..end].to_vec()
+    payload.to_vec()
 }
--- a/eng-wasm/src/crypto_wasm.rs
+++ b/eng-wasm/src/crypto_wasm.rs
@@ -1,14 +1,18 @@
+/// Wrapper for the Enigma runtime service for symmetric ASM-256-GCM encryption/decryption
+/// The encrypted text contains the IV, cyphertext, and authentication tag
+/// In AES-GCM the length of cyphertext is identical to the length of plain text
+
 use super::*;
 
 use eng_pwasm_abi::types::U256;
 use rand_wasm::Rand;
 
-const SYMMETRIC_KEY_SIZE: usize = 32;
-
+const SYMMETRIC_KEY_SIZE: usize = 256 / 8;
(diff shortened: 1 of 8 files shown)


moriaab:
@reuvenpo Please review. This is not a complete PR. After your review I will push the new version of `eng-wasm`, then will change the dependency in  `examples/eng_wasm_contracts/encryption/Cargo.toml` to `crates.io` again. The new version of `eng-wasm` will not be backward compatible, therefore for all wasm examples I will change the dependencies in `Cargo.toml` to `eng-wasm = "0.2"`.

moriaab:
@Cashmaney 
This PR fixes issue https://github.com/enigmampc/enigma-core/issues/230. The changes in `eng-wasm/src/crypto_wasm.rs` are the most relevant ones for the issue. Please pay attention that we use AES in GCM mode, therefore the size of the cypher text is:
plain text size + tag size. 
(cc @reuvenpo)

enigma-types/src/types.rs
```diff
@@ -12,8 +12,10 @@
 use core::{fmt, mem, ptr, default::Default};
 
 pub use crate::hash::Hash256;
-/// SymmetricKey is the 256bit symmetric key we use for encryption.
-pub type SymmetricKey = [u8; 32];
+/// The size of the symmetric 256 bit key we use for encryption (in bytes).
+pub const SYMMETRIC_KEY_SIZE: usize = 32;
```
reuvenpo: I won't be too picky about it, but it's clearer to write this as the following expression:
```rust
/// The size of the symmetric 256 bit key we use for encryption (in bytes).
pub const SYMMETRIC_KEY_SIZE: usize = 256 / 8;  // 8 bits to a byte
```
This expresses the intent better IMO

eng-wasm/src/lib.rs
```diff
@@ -37,7 +37,10 @@ pub mod external {
         pub fn rand(payload: *const u8, payload_len: u32);
         pub fn encrypt(message: *const u8, message_len: u32, key: *const u8, payload: *const u8);
         pub fn decrypt(cipheriv: *const u8, cipheriv_len: u32, key: *const u8, payload: *const u8);
-    }
+        pub fn get_iv_size() -> u32;
+        pub fn get_tag_size() -> u32;
+
+}
```
reuvenpo: add indentation to this `}`

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
 use rand_wasm::Rand;
 
 const SYMMETRIC_KEY_SIZE: usize = 32;
-const IV_SIZE: usize = 96 / 8;
-const PAYLOAD_SIZE: usize = 1024;
 
-pub type IV = [u8; IV_SIZE];
 pub type SymmetricKey = [u8; SYMMETRIC_KEY_SIZE];
 
+fn extra_size_for_cypher() -> usize {
```
Cashmaney: cipher, not cypher:)

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
...
-            // Range selectors exclude the upper bound
-            end = i + 1;
-            break;
-        }
-    }
-    payload[0..end].to_vec()
+    payload.to_vec()
 }
 
 pub fn decrypt(cipheriv: &[u8], key: &SymmetricKey) -> Vec<u8> {
```
Cashmaney, line 38: Wait, why is this called cipheriv? Isn't this the encrypted message?

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
...
-    // Assuming that plaintext messages cannot be shorter that their encrypted cipher
-    // TODO: Some unnecessary bytes can be subtracted from the buffer, at least the IV size.
-    let length: usize = cipheriv.len();
-    let mut payload = Vec::with_capacity(length);
-    for _ in 0..length {
-        payload.push(0);
-    }
+    let length = cipheriv.len() - extra_size_for_cypher();
```
Cashmaney: Are we sure this works? If we have 32 bytes that are supposed to be decrypted to 32 bytes it seems odd to me that we're subtracting length here

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
...
 
 const SYMMETRIC_KEY_SIZE: usize = 32;
-const IV_SIZE: usize = 96 / 8;
-const PAYLOAD_SIZE: usize = 1024;
 
-pub type IV = [u8; IV_SIZE];
 pub type SymmetricKey = [u8; SYMMETRIC_KEY_SIZE];
 
+fn extra_size_for_cypher() -> usize {
+    unsafe {(external::get_iv_size() + external:: get_tag_size()) as usize }
```
Cashmaney: Should comment what this tag_size is and why you need it

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
...
 
 pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    // TODO: Is this really needed? Dynamically sized buffers don't seem to work.
-    // TODO: Is it possible to estimate the encrypted payload size based on the plaintext message?
-    let length = PAYLOAD_SIZE;
-    let mut payload = Vec::with_capacity(length);
-    for _ in 0..length {
-        payload.push(0);
-    }
+    let length = message.len() + extra_size_for_cypher();
```
Cashmaney: Are we sure this is also working? If extra_size..() is constant, but message.len() is obviously dynamic, how can we be sure we allocate the exact size of the encrypted message?

Cashmaney (changes requested):

eng-wasm/src/crypto_wasm.rs
Cashmaney: Ah, because it is the ciphertext + iv? =/

enigma-crypto/src/symmetric.rs
```diff
@@ -74,6 +74,12 @@ pub fn decrypt(cipheriv: &[u8], key: &SymmetricKey) -> Result<Vec<u8>, CryptoErr
     Ok(decrypted_data.to_vec())
 }
 
+/// Get tag size
+pub fn get_tag_size() -> usize {
+    AES_MODE.tag_len()
+}
```
reuvenpo: please add an `#[inline(always)]` annotation to this, to reduce the runtime cost. `AES_MODE.tag_len()` (which is really `ring::aead::Algorithm::tag_len()`) is also defined with this, because all it does is return a constant from inside the `ring::aead` scope.

I understand the maintainability consideration in moving this value out of eng_wasm and into the runtime, but is the runtime overhead of context-switching from WASM to Rust really worth all this trouble?

(also, how is this relevant to the main task? I must be missing something)

eng-wasm/src/crypto_wasm.rs
```diff
@@ -17,41 +16,15 @@ pub fn generate_key() -> SymmetricKey {
...
 
 pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    // TODO: Is this really needed? Dynamically sized buffers don't seem to work.
-    // TODO: Is it possible to estimate the encrypted payload size based on the plaintext message?
-    let length = PAYLOAD_SIZE;
-    let mut payload = Vec::with_capacity(length);
-    for _ in 0..length {
-        payload.push(0);
-    }
+    let length = message.len() + IV_SIZE + TAG_SIZE;
```
reuvenpo: Shouldn't you use `extra_size_for_cypher` instead of `IV_SIZE + TAG_SIZE` here?
Also, this is not the calculation that itzik demostrated in #230. Can you please explain this new calculation?

eng-wasm/src/crypto_wasm.rs
```diff
@@ -17,41 +16,15 @@ pub fn generate_key() -> SymmetricKey {
...
 
 pub fn decrypt(cipheriv: &[u8], key: &SymmetricKey) -> Vec<u8> {
-    // Assuming that plaintext messages cannot be shorter that their encrypted cipher
-    // TODO: Some unnecessary bytes can be subtracted from the buffer, at least the IV size.
-    let length: usize = cipheriv.len();
-    let mut payload = Vec::with_capacity(length);
-    for _ in 0..length {
-        payload.push(0);
-    }
+    let length: usize = cipheriv.len() - IV_SIZE - TAG_SIZE;
```
reuvenpo: same comments as above in `fn encrypt`.
you can write this as:
```rust
let length: usize = cipheriv.len() - extra_size_for_cypher();
```

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
...
+}
+
+
 pub fn generate_key() -> SymmetricKey {
     let key_int: U256 = Rand::gen();
     let key = H256::from(key_int);
     key.0
 }
 
 pub fn encrypt(message: &[u8], key: &SymmetricKey) -> Vec<u8> {
```
Cashmaney, line 28: Should add a little bit of text, or at least a reference as to why this works the way it does

eng-wasm/src/internal_std.rs
```diff
@@ -8,6 +8,7 @@ pub use self::std::{
     str::from_utf8,
     string::{String, ToString},
     vec::Vec,
+    vec,
```
reuvenpo: Note that this exposes the entire `vec` namespace, not just the `vec!` macro.
If this is not ok with you, you can instead use `#[macro_use(vec)]`.
[This is explained here](https://doc.rust-lang.org/reference/macros-by-example.html#the-macro_use-attribute).

reuvenpo:
As a general comment, it really looks like we can easily compress the code-size and maintainability overhead of `enigma-runtime-t` by refactoring about half of it as a macro_rules macro. Obviously not a priority right now, but worth keeping in mind.

Also @Cashmaney asked me to mention that we should add a check in the `encrypt` and `decrypt` functions that the arithmetic there doesn't overflow/underflow the integers. (check out the `checke_{add,sub}` methods in the standard library)

eng-wasm/src/crypto_wasm.rs
```diff
@@ -4,54 +4,30 @@ use eng_pwasm_abi::types::U256;
...
     unsafe { external::decrypt(cipheriv.as_ptr(), cipheriv.len() as u32, key.as_ptr(), payload.as_mut_ptr()) };
-    let mut end = 0;
-    for i in (0..length).rev() {
-        if payload[i] != 0 {
-            end = i + 1;
-            break;
-        }
-    }
-    payload[0..end].to_vec()
+    payload.to_vec()
```
reuvenpo: why are you using `to_vec` here? it's already a vector

eng-wasm/src/crypto_wasm.rs
moriaab: @Cashmaney 
From wikipedia:
In cryptography, a cipher (or *cypher*) is an algorithm for performing encryption or decryption...

eng-wasm/src/crypto_wasm.rs
moriaab: yes

eng-wasm/src/crypto_wasm.rs
moriaab: Done

enigma-types/src/types.rs
moriaab: I agree. Done.

eng-wasm/src/lib.rs
moriaab: Done

eng-wasm/src/crypto_wasm.rs
moriaab: Done

eng-wasm/src/crypto_wasm.rs
moriaab: Good catch. It was in the previous implementation. Didn't pay attention.

enigma-crypto/src/symmetric.rs
moriaab: @reuvenpo I understand your concern about runtime overhead, I implemented this for maintainability, but never was sure that this worth the overhead. Will be happy to discuss this f2f.
I don't understand your comment about "main task". What is main task?

eng-wasm/src/internal_std.rs
moriaab: Done (with your help :))

moriaab:
@reuvenpo The last comment regarding overflow is addressed.

enigma-crypto/src/symmetric.rs
moriaab: As we decided, I define the constants in `crypto_wasm.rs` to avoid runtime overhead

reuvenpo (approved):

Cashmaney (approved):

PR #243, opened 2019-10-28
configurable epoch cap size

AvishaiW:
added as requested by @lenak25. 
This PR makes the epoch cap size configurable as opposed to what was until now where we had it hardcoded on a constant.
the epoch cap size tells us how many epochs to store in memory at any given time.


AvishaiW:
waiting for [this PR](https://github.com/enigmampc/discovery-docker-network/pull/44) to be added which will enable us to run the CI.

PR #246, opened 2019-10-31
New block_number definition to distinguish between km & Ether block_number

AvishaiW:
This PR fixes a bug that was found when running a network with many nodes which increases the latency and which simulates a realistic network.

The KM defines the new epoch on a block and uses it to get the list of the workers which will be participating in the following epoch.
Then it sends the Event to the Enigma Contract (**which might take time to get it**) and it signs a transaction which defines the **real** start of the epoch and might be in a different block_number.

Till now, we stored in the enclave of the km the km_block_number and in the untrusted we stored the ether_block_number and when they needed to be verified one against the other bad things happened.

So we split it into two numbers in the untrusted side, and when needed to know the real epoch the ether_block_number is given and when needed to verify the workers participating in the epoch the km_block_number is given.

### DIFF
--- a/enigma-principal/app/src/epoch_u/epoch_provider.rs
+++ b/enigma-principal/app/src/epoch_u/epoch_provider.rs
@@ -125,7 +125,7 @@ impl EpochStateManager {
         let epoch_states = self.get_all_confirmed()?;
         for epoch_state in epoch_states.iter() {
             let confirmed: &ConfirmedEpochState = epoch_state.confirmed_state.as_ref().unwrap();
-            if confirmed.block_number <= block_number {
+            if confirmed.ether_block_number <= block_number {
                 result = Some(epoch_state);
             }
         }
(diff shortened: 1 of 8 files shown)


enigma-principal/app/src/epoch_u/epoch_types.rs
```diff
@@ -17,19 +17,27 @@ pub const WORKER_PARAMETERIZED_EVENT: &str = "WorkersParameterized";
...
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct EpochState {
     pub seed: U256,
     pub sig: Bytes,
     pub nonce: U256,
+    /// The km_block_number is the block in which the KM decided to start a new epoch and
+    /// the active workers are concluded from for the epoch
+    /// (It might differ from the ether_block_number due to latency in networks)
+    pub km_block_number: U256,
```
moriaab, line 33: Excellent explanation!

enigma-principal/app/src/epoch_u/epoch_provider.rs
```diff
@@ -318,21 +320,28 @@ impl EpochProvider {
     }
 
     #[logfn(DEBUG)]
-    fn set_worker_params_internal<G: Into<U256>>(&self, block_number: U256, gas_limit: G, confirmations: usize, epoch_state: Option<EpochState>) -> Result<H256, Error> {
-        let result = self.contract.get_active_workers(block_number)?;
-        let worker_params = InputWorkerParams { block_number, workers: result.0, stakes: result.1 };
+    fn set_worker_params_internal<G: Into<U256>>(&self, km_block_number: U256, gas_limit: G, confirmations: usize, epoch_state: Option<EpochState>) -> Result<H256, Error> {
+        let (workers, stakes) = self.contract.get_active_workers(km_block_number)?;
```
moriaab, line 324: Well done! Now it is much more clear.

enigma-principal/app/src/epoch_u/epoch_types.rs
```diff
@@ -39,7 +47,7 @@ impl EpochState {
     /// * `sc_addresses` - The Secret Contract addresses for which to retrieve the selected worker
     #[logfn(DEBUG)]
     pub fn confirm(
-        &mut self, block_number: U256, worker_params: &InputWorkerParams, sc_addresses: Vec<ContractAddress>,
+        &mut self, ether_block_number: U256, worker_params: &InputWorkerParams, sc_addresses: Vec<ContractAddress>,
```
moriaab, line 50: Should this function use `println!` or `info!`?

moriaab (approved):

PR #247, opened 2019-11-13
added local signing of calls to `register` and `setWorkersParams` in `enigma-tools-u`

reuvenpo:
Please note that the private key and the chain id are currently hardcoded to `[0_u8; 32]` and `1` respectively, and should be taken from the configuration instead. This means that this branch will definitely fail in CI until we polish this.

EDIT: these values are now read from configuration, but the chain id is truncated to u8 because of a bug in ethereum-tx-sign.


enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+
+from_error!(Error::IoError, std::io::Error);
+from_error!(Error::SerdeJsonError, serde_json::Error);
+from_error!(Error::ParseIntError, std::num::ParseIntError);
+
+fn option_to_u256(value: Option<U256>) -> ethereum_types::U256 {
+    value.unwrap_or_else(|| ethereum_types::U256::from(0))
+}
+
+/// Confused? You should be. Please read the documentation in the top of this file.
```
reuvenpo, line 99: I should add a comment that this is based on [this](https://tomusdrw.github.io/rust-web3/web3/contract/struct.Contract.html#method.call_with_confirmations)

enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+    IoError(#[cause] std::io::Error),
+
+    #[fail(display = "{}", _0)]
+    SerdeJsonError(#[cause] serde_json::Error),
+
+    #[fail(display = "{}", _0)]
+    ParseIntError(#[cause] std::num::ParseIntError),
+}
+
+macro_rules! from_error_to_string {
```
reuvenpo, line 67: I Should add a comment that some error types can not be used with `failure` because they don't implement `(dyn std::error::Error + Send + Sync + 'static)`, and that's why we convert them to `String`s

enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+        H256,
+        BlockNumber,
+    },
+};
+
+/// I implement this struct to coalesce all the errors from the crates used in this file into one struct, that can
+/// be `?`-cast into `failure::Fail`
+#[derive(Debug, failure::Fail)]
+pub enum Error {
+    #[fail(display = "{}", _0)]
```
AvishaiW, line 46: what does `_0` mean?

enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+}
+
+from_error!(Error::IoError, std::io::Error);
+from_error!(Error::SerdeJsonError, serde_json::Error);
+from_error!(Error::ParseIntError, std::num::ParseIntError);
+
+fn option_to_u256(value: Option<U256>) -> ethereum_types::U256 {
+    value.unwrap_or_else(|| ethereum_types::U256::from(0))
```
AvishaiW:
```suggestion
    value.unwrap_or_else(|| ethereum_types::U256::zero())
```

enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+/// Confused? You should be. Please read the documentation in the top of this file.
+pub fn signed_call_with_confirmations<T: Transport, P>(
+    // general tools
+    web3: &web3::api::Web3<T>,
+    // contract details
+    contract_abi: &ethabi::Contract,
+    contract_address: Address,
+    // sender details
+    private_key: H256,
+    from: Address, // Should belong to the private key
```
AvishaiW, line 111: how do you verify there is a connection between the address and the priv key? (@moriaab )

enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+        gas: option_to_u256(options.gas),
+        data: fn_data,
+    };
+
+    let signed_tx = tx.sign(&private_key, &chain_id);
+
+    Ok(confirm::send_raw_transaction_with_confirmation(
+        web3.eth().transport().clone(),
+        Bytes::from(signed_tx),
+        poll_interval,
```
AvishaiW, line 148: what is the purpose of this interval and why is it of 1 sec?

enigma-tools-u/src/web3_utils/enigma_contract.rs
```diff
@@ -23,54 +24,38 @@ pub struct EnigmaContract {
     pub web3: Arc<Web3<Http>>,
     pub eloop: EventLoopHandle,
     pub w3_contract: Contract<Http>,
```
AvishaiW, line 26: is this still necessary here? seems like it's needed only for the address that I'm guessing can be extracted from the ethabi contract as well?

enigma-tools-u/src/web3_utils/enigma_contract.rs
```diff
@@ -23,54 +24,38 @@ pub struct EnigmaContract {
...
+        contract_address: &str,
+        abi_path: P,
+        account: Option<&str>,
+        web3: Web3<Http>,
+        eloop: EventLoopHandle
+    ) -> Result<Self, Error> {
         let account: Address = match account {
             Some(a) => a.parse()?,
-            None => web3.eth().accounts().wait().unwrap()[0], // TODO: Do something with this unwrapping
+            None => web3.eth().accounts().wait()?[0],
```
AvishaiW, line 59: Great!

enigma-tools-u/src/web3_utils/enigma_contract.rs
```diff
@@ -88,35 +73,53 @@ pub trait ContractFuncs<G> {
...
-                return Err(errors::Web3Error { message: format!("Unable to call register: {:?}", e) }.into());
-            }
-        };
+        let report = report.as_bytes().to_vec();
+        let signature = signature.from_hex()?;
+        let call = signed_call_with_confirmations(
+            &self.web3,
+            &self.ethabi_contract,
+            self.w3_contract.address(),
+            H256([0; 32]), // TODO set to a real key
```
AvishaiW: make sure not to push this to develop! (I know you are working on this)

enigma-tools-u/src/web3_utils/contract_ext.rs
```diff
@@ -0,0 +1,146 @@
...
+    contract_address: Address,
+    // sender details
+    private_key: H256,
+    from: Address, // Should belong to the private key
+    // function details
+    func: &str,
+    params: P,
+    // network details
+    options: Options,
+    chain_id: u8,
```
AvishaiW, line 117: maybe add somewhere an explanation about the chain_id. Thanks!

AvishaiW (changes requested):
looking good!
waiting for you to push the changes regarding the configured priv key & chain_id

enigma-tools-u/src/web3_utils/contract_ext.rs
reuvenpo: It's referring to the first(and only, in this case) item in the tuple-struct variant `Web3Error`, which is a `String`. That is then fed into the `{}` of the format string.`

enigma-tools-u/src/web3_utils/contract_ext.rs
reuvenpo: Good idea! Thanks :)

enigma-tools-u/src/web3_utils/contract_ext.rs
reuvenpo: I currently do not. ideally we would generate the address from this key, but i didn't want to deal with that at this point. We will probably add this before this gets merged

enigma-tools-u/src/web3_utils/contract_ext.rs
AvishaiW: you can add it immediately if you press on `commit suggestion`

enigma-tools-u/src/web3_utils/contract_ext.rs
reuvenpo: This is the same interval used in the original implementation [here](https://github.com/tomusdrw/rust-web3/blob/c69bf938a0d3cfb5b64fca5974829408460e6685/src/contract/mod.rs#L153)

enigma-tools-u/src/web3_utils/enigma_contract.rs
reuvenpo: Good point, i'll look into it.

enigma-tools-u/src/web3_utils/enigma_contract.rs
reuvenpo: Thanks :)

enigma-tools-u/src/web3_utils/contract_ext.rs
reuvenpo: I added it in the commits you can't see yet.

enigma-tools-u/src/web3_utils/enigma_contract.rs
reuvenpo: Thanks!

enigma-tools-u/src/web3_utils/contract_ext.rs
reuvenpo: I know, but i prefer to organize all these changes manually :+1:

enigma-tools-u/src/web3_utils/enigma_contract.rs
reuvenpo: nope, we also use its `query` method under `enigma-tools-u/web3_utils::enigma_contract::<impl ContractQueries for EnigmaContract>`

AvishaiW (approved):
Great Job and thanks for removing all the dead code! (for next time, let's split this into 2 PR's)

enigma-tools-u/src/web3_utils/contract_ext.rs
AvishaiW: maybe add a TODO then?

reuvenpo: I added this task to our internal backlog

reuvenpo:
This PR is superseded by PR #252 which built on it and expanded its functionality. please refer to the discussion there.

PR #254, opened 2019-11-28
Health checks

AvishaiW:
Added a Health Check for the KM.
This is used to check if the KM is up and running from the Kubernetes cluster side.
call it when the km is up using: 
`curl -sb -o /dev/null -X POST -d '{"jsonrpc": "2.0", "id": "1", "method": "getHealthCheck", "params": []}' -H "Content-Type: application/json" km:3040`


enigma-principal/app/src/boot_network/keys_provider_http.rs
```diff
@@ -168,18 +171,39 @@ impl PrincipalHttpServer {
...
+        // Ethereum
+        let contract_signing_address = match epoch_provider.contract.get_signing_address() {
+            Ok(addr) => addr,
+            Err(_) => return Value::Bool(false),
+        };
+        // Enclave
+        let enclave_signing_address: H160 = match esgx::equote::get_register_signing_address(*epoch_provider.eid) {
+            Ok(addr) => addr.into(),
+            Err(_) => return Value::Bool(false),
+        };
```
moriaab, line 187: It is cleaner to use `unwrap_or_else(||return Value:Bool(false))` in both `let` expressions.

moriaab (approved):
LGTM

enigma-principal/app/src/boot_network/keys_provider_http.rs
AvishaiW: @moriaab I thought so as well (notice that I've done that in the first commits of this PR) but notice that the return is for the function inside the `unwrap_or_else` and it will be assigned to `enclave_signing_address` leading to an error (the types aren't aligned).

PR #256, opened 2019-12-02
Introducing ISVPRODID

moriaab:
ISVPRODID is the way for enclave provider to differentiate between different kinds of enclaves, such that they will not be able to unseal information of one another. In our case, there are two enclaves: `KM` and `core`.


AvishaiW (approved):
Love the new numbers!

PR #259, opened 2019-12-09
fixed register call according to the change in the enigma contract

AvishaiW:
https://github.com/enigmampc/enigma-contract/pull/177
since it's irrelevant for the KM I added an empty string in the staking address field


moriaab (approved):
LGTM

PR #261, opened 2019-12-15
Update Cargo.toml

Cashmaney:


AvishaiW (approved):

AvishaiW:
I'm debating if it would be smart to add the feature in the enigma-principal as well

PR #268, opened 2019-12-23
Changed retrieval of bytecodes to bytes instead of strings- IPC

AvishaiW:


enigma-core/app/tests/ipc_computation_tests.rs
```diff
@@ -101,8 +101,7 @@ fn test_execute_on_existing_contract_with_constructor() {
     let computed_data: Vec<u8> = serde_json::from_value(add_delta["data"].clone()).unwrap();
 
     let new_addr = generate_contract_address();
-    let _msg = get_msg_format_update_contract(&new_addr.to_hex(), deployed_bytecode);
-    let _res_a = send_update_contract(port, &new_addr.to_hex(), deployed_bytecode);
+    let _res_a = send_update_contract(port, &new_addr.to_hex(), deployed_bytecode.from_hex().unwrap());
```
reuvenpo, line 103: Neither of these are used. i suspect it's here to make sure it doesn't panic. you should return the line you removed.

AvishaiW: The line I removed appears inside the send_update_contract method- most probably it was never removed from here by mistake. Thanks!

reuvenpo (commented):
looks good :)

reuvenpo (approved):

PR #269, opened 2019-12-23
FIX: fixing corner case in factorization secret contract

moriaab:
This PR fixes bug in factorization secret contract: when a number has prime factor greater than its sqrt, it was not counted.

### DIFF
--- a/enigma-core/app/src/wasm_u/wasm.rs
+++ b/enigma-core/app/src/wasm_u/wasm.rs
@@ -380,7 +380,7 @@ mod tests {
         let address = generate_contract_address();
 
         // All prime factors are less than sqrt of an input
-        let (enclave, contract_code, result, shared_key) = compile_deploy_execute(
+        let (enclave, contract_code, result1, shared_key1) = compile_deploy_execute(
             &mut db,
             "../../examples/eng_wasm_contracts/factorization",
             address,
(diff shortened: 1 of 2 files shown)


examples/eng_wasm_contracts/factorization/src/lib.rs
```diff
@@ -19,6 +19,9 @@ impl S {
                 number = number / i;
```
AvishaiW, line 20: I noticed that on square root odd numbers that are prime numbers as well (say 7 for 49, 11 for 121 etc.) will not return the correct answer since the range doesn't cover it. I think you need to add one to the higher bound of the range.

enigma-core/app/src/wasm_u/wasm.rs
```diff
@@ -388,14 +389,15 @@ mod tests {
             "find_number_of_prime_factors(uint32)",
             &[Token::Uint(199998.into())]
         );
-        let encoded_output = symmetric::decrypt(&result.output, &shared_key).unwrap();
+        let mut encoded_output = symmetric::decrypt(&result.output, &shared_key).unwrap();
```
AvishaiW: it seems that the variables are used for different purposes- IMO that should have different naming for each section, otherwise, it can cause confusion in the future.

examples/eng_wasm_contracts/factorization/src/lib.rs
AvishaiW: another edge case: 0 can't be [right-shifted](https://github.com/enigmampc/enigma-core/pull/269/files#diff-d3895e8dde647512c064d1e384c0e0f1L14) and therefore, returns an error.

AvishaiW (changes requested):

examples/eng_wasm_contracts/factorization/src/lib.rs
moriaab: Excellent catch! Thank you!

enigma-core/app/src/wasm_u/wasm.rs
moriaab: Ok. Will fix this.

AvishaiW (approved):
Excellent!

PR #272, opened 2019-12-31
added clarity to DB logs

AvishaiW:
added clarity to the error logs coming from the DB. get_deltas and get_contract would get until now a "read" error so I modified it to the function which made the call and in addition, I added the contract address of which we are querying.

### DIFF
--- a/enigma-core/app/src/common_u/errors.rs
+++ b/enigma-core/app/src/common_u/errors.rs
@@ -59,18 +59,20 @@ pub enum DBErrKind {
     KeyExists,
     CreateError,
     FetchError,
-    MissingKey,
+    MissingKey(String),
     UpdateError,
+    MissingKeys,
 }
 
 impl fmt::Display for DBErrKind {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let printable = match *self {
-            DBErrKind::KeyExists => "The Key already exists",
-            DBErrKind::CreateError => "Failed to create the key",
-            DBErrKind::FetchError => "Failed to fetch the data",
-            DBErrKind::MissingKey => "The Key doesn't exist",
-            DBErrKind::UpdateError => "Failed to update the key",
+        let printable: String = match &*self {
+            DBErrKind::KeyExists => "The Key already exists".into(),
+            DBErrKind::CreateError => "Failed to create the key".into(),
+            DBErrKind::FetchError => "Failed to fetch the data".into(),
+            DBErrKind::MissingKey(k) => format!("The following Key doesn't exist: {}", &k),
+            DBErrKind::UpdateError => "Failed to update the key".into(),
+            DBErrKind::MissingKeys => "Keys don't exist the DB".into(),
         };
         write!(f, "{}", printable)
     }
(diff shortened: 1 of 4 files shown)


enigma-core/app/src/common_u/errors.rs
```diff
@@ -59,18 +59,20 @@ pub enum DBErrKind {
...
-            DBErrKind::FetchError => "Failed to fetch the data",
-            DBErrKind::MissingKey => "The Key doesn't exist",
-            DBErrKind::UpdateError => "Failed to update the key",
+        let printable: String = match &*self {
+            DBErrKind::KeyExists => "The Key already exists".into(),
+            DBErrKind::CreateError => "Failed to create the key".into(),
+            DBErrKind::FetchError => "Failed to fetch the data".into(),
+            DBErrKind::MissingKey(k) => format!("The following Key doesn't exist: {}", &k),
+            DBErrKind::UpdateError => "Failed to update the key".into(),
+            DBErrKind::MissingKeys => "Keys don't exist the DB".into(),
```
moriaab: I think that the message in not so clear. Should be changed to something like: "No keys in DB".

enigma-core/app/src/db/dal.rs
```diff
@@ -216,8 +216,8 @@ impl<'a, K: SplitKey> CRUDInterface<Error, &'a K, Vec<u8>, &'a [u8]> for DB {
     fn read(&self, key: &'a K) -> Result<Vec<u8>, Error> {
         key.as_split(|hash, index_key| {
             trace!("DB: Read: contract_address: {}, key: {:?}", hash, index_key);
-            let cf_key = self.database.cf_handle(&hash).ok_or(DBErr { command: "read".to_string(), kind: DBErrKind::MissingKey })?;
-            let value = self.database.get_cf(cf_key, &index_key)?.ok_or(DBErr { command: "read".to_string(), kind: DBErrKind::MissingKey })?;
+            let cf_key = self.database.cf_handle(&hash).ok_or(DBErr { command: "read".to_string(), kind: DBErrKind::MissingKey(hash.to_string()) })?;
```
moriaab: I think that the same change should be done for `create` in line 204.

enigma-core/app/src/common_u/errors.rs
moriaab, line 75: I think that the message in not so clear. Should be changed to something like: "No keys in DB".

enigma-core/app/src/common_u/errors.rs
```diff
@@ -59,18 +59,20 @@ pub enum DBErrKind {
     KeyExists,
```
moriaab, line 59: I think that `KeyExists` should contain the key as `MissingKey`.

enigma-core/app/src/db/iterator.rs
```diff
@@ -348,7 +350,7 @@ impl P2PCalls for DB {
         from.as_split(|from_hash, from_key| {
             // make sure the address exists as a CF in the DB
             let cf_key =
-                self.database.cf_handle(&from_hash).ok_or(DBErr { command: "read".to_string(), kind: DBErrKind::MissingKey })?;
+                self.database.cf_handle(&from_hash).ok_or(DBErr { command: "read".to_string(), kind: DBErrKind::MissingKey(from_hash.to_string()) })?;
```
moriaab, line 353: Shouldn't `get_deltas` be used as `command` here (instead of `read`)?

moriaab (approved):
Good work!

PR #274, opened 2020-01-02
MAINT: Cargo.toml not executable

lacabra:
Some `Cargo.toml` were marked executable, but they don't need to be. They are regular files. This PR fixes that.


AvishaiW (approved):
Thanks! did this affect the network in some sort?

lacabra:
No, it does not affect the network in any way. I just came across this when pulling the latest version of the factorization contract.

PR #281, opened 2020-01-19
fixes DB recreation after failure

AvishaiW:
This PR makes sure the options assigned to the DB are stored, and it recreates the DB as expected.
this was fixed according to [this issue](https://github.com/rust-rocksdb/rust-rocksdb/issues/370#issuecomment-573399033)

### DIFF
--- a/enigma-core/app/src/db/dal.rs
+++ b/enigma-core/app/src/db/dal.rs
@@ -1,6 +1,6 @@
 use failure::Error;
 use rocksdb::DB as rocks_db;
-use rocksdb::{Options, SliceTransform, WriteOptions};
+use rocksdb::{Options, SliceTransform, WriteOptions, ColumnFamilyDescriptor};
 use std::path::{Path, PathBuf};
 
 use common_u::errors::{DBErr, DBErrKind};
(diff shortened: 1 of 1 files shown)


enigma-core/app/src/db/dal.rs
```diff
@@ -51,9 +51,14 @@ impl DB {
             Ok(list) => list,
             Err(_) => Vec::new(),
         };
-        // converts the Strings to slices (str)
-        let cf_list_burrowed = cf_list.iter().map(String::as_str).collect::<Vec<&str>>();
-        let database = rocks_db::open_cf(&options, &location, &cf_list_burrowed[..])?;
+        // converts the Strings to descriptors (adds to each cf an options object)
+        let cf_descriptors = cf_list.into_iter().map(|name| {
+            let prefix_extractor = SliceTransform::create_fixed_prefix(1);
```
moriaab: It is better to use `DELTA_PREFIX` here and in line 42 instead of a hard-coded value `1`.

moriaab (approved):
What a relief that you found a solution!

PR #283, opened 2020-01-27
fixing ncc 009

AvishaiW:
added a condition when serializing a msg into PrincipalMessage.

I noticed that we duplicate the code which I'm not sure why we do that. needed to be checked in a different PR.

### DIFF
--- a/enigma-tools-m/src/primitives/km_primitives.rs
+++ b/enigma-tools-m/src/primitives/km_primitives.rs
@@ -13,6 +13,18 @@ use enigma_types::{ContractAddress, DhKey, PubKey, StateKey};
 /// A Message ID type, used to identify each message to the response.
 pub type MsgID = [u8; 12];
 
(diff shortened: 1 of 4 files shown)


moriaab (approved):
Although the fix of NCC issue `009` is correct, I would suggest another implementation:
1. Define constant for pub key size in `enigma-types/src/types.rs` (for example `PUB_KEY_SIZE`) and use this constant instead if `64` in the definition `pub type PubKey = [u8; 64];`
2. In `impl PrincipalMessage` and `impl UserMessage` add function 
```
fun verify_key_size() ->Result<(), ToolsError> {
if msg.pubkey.len() != PUB_KEY_SIZE {
            Err(MessagingError { err: "the pub key is not of the right length" })
}
OK(())
}
```
3. Use `verify_key_size` in: `from_message` `into_message` and `get_pubkey`

I think that the correct message is "the pub key is not **of** the right length"

PR #284, opened 2020-01-29
Rolling log file

AvishaiW:
added a rolling log file with 3 archived files with a size of 300 MB each.
added an ignored test which can check that rolling is happening- only necessary for changes in the log config

### DIFF
--- a/enigma-tools-u/src/common_u/logging.rs
+++ b/enigma-tools-u/src/common_u/logging.rs
@@ -12,18 +12,19 @@ use failure::Error;
 
 use log::{LevelFilter};
 use log4rs::append::console::{ConsoleAppender, Target};
-use log4rs::append::file::FileAppender;
+use log4rs::append::{file::FileAppender, rolling_file::RollingFileAppender};
 use log4rs::config::{Appender, Config, Root, Logger};
 use log4rs::encode::pattern::PatternEncoder;
 use log4rs::filter::threshold::ThresholdFilter;
 use log4rs::Handle;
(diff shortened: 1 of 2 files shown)


enigma-tools-u/src/common_u/logging.rs
```diff
@@ -33,11 +37,27 @@ pub fn init_logger<P: AsRef<Path>>(level: log::LevelFilter, data_dir: P, name: S
...
         .target(Target::Stdout).build();
 
-    // Logging to log file.
-    let logfile = FileAppender::builder()
+    // define how many archived logs there will be aside from the log that will be written into.
+    // see docs here:
+    // https://docs.rs/log4rs/0.9.0/log4rs/append/rolling_file/policy/compound/roll/fixed_window/struct.FixedWindowRoller.html
+    // to understand how it works.
+    // see this issue as well:
+    // https://github.com/estk/log4rs/issues/120
```
moriaab, line 45: Thanks for the documentation!

enigma-tools-u/src/common_u/logging.rs
```diff
@@ -73,3 +86,25 @@ pub fn init_logger<P: AsRef<Path>>(level: log::LevelFilter, data_dir: P, name: S
     let handle = log4rs::init_config(config).unwrap();
     Ok(handle)
 }
+
+
+#[cfg(test)]
```
moriaab: As we discussed, you may remove this test since it does not test our functionality.

moriaab (approved):
Well done on understanding `log4rs` complex API and documenting the code!

PR #285, opened 2020-01-30
MAINT: max_epochs=0

lacabra:
CIs started picking up `max_epochs=10` from the default config, so this PR resets it to `0`


lacabra:
This is arguably redundant since the issue that affects the CIs has been fixed already in enigmampc/docker-environment#90, but for most intents and purposes this value should be `0`, it is preferable to set the default to `0`.

AvishaiW:
I'm actually working on refactoring significant parts of the KM at the moment and hopefully I'll be removing this as well.

AvishaiW (approved):

